PR #63, opened 2020-10-30
feat: selectors

yury-s:


playwright/src/main/java/com/microsoft/playwright/impl/ElementHandleImpl.java
```diff
@@ -384,6 +384,16 @@ private static String toProtocol(ElementState state) {
     return toDeferred(sendMessageAsync("waitForElementState", params).apply(json -> null));
   }
 
+  public String createSelectorForTest(String name) {
```
kblok, line 387: I’d rather ignore the test than making this part of the public API.

yury-s: This is not getting into the public interface, only to the impl class (the test has to cast `ElementHandle` to `ElementHandleImpl` to use it) . If clients want to abuse the implementation they can already do this even for package private methods - just declare their own classes in the same package.

PR #71, opened 2020-11-06
docs: add javadoc comments to interfaces and methods

yury-s:


playwright/src/main/java/com/microsoft/playwright/Accessibility.java
```diff
@@ -35,6 +46,28 @@ public SnapshotOptions withRoot(ElementHandle root) {
   default AccessibilityNode snapshot() {
     return snapshot(null);
   }
+  /**
+   * Captures the current state of the accessibility tree. The returned object represents the root accessible node of the page.
+   * <p>
+   * 
+   * <p>
```
kblok, line 53: Maybe ignore empty lines?

playwright/src/main/java/com/microsoft/playwright/Accessibility.java
```diff
@@ -35,6 +46,28 @@ public SnapshotOptions withRoot(ElementHandle root) {
...
+   * An example of dumping the entire accessibility tree:
+   * <p>
+   * 
+   * <p>
+   * An example of logging the focused node's name:
+   * <p>
+   *
```
kblok, line 68: What would be the plan here?

yury-s: Don't know yet, we'll have to figure out a story that would work across languages and be manageable in terms of long term maintenance. We brainstormed a few approaches like embedding language-specific snippets into api.md or approach documentation similar to localization strings but all of those had problems that we didn't like. For now I'm just removing the snippets to avoid confusion. I can try and remove the lines starting with 'An example of' too.

playwright/src/main/java/com/microsoft/playwright/Accessibility.java
yury-s: Lemme try that.

PR #90, opened 2020-12-06
devops: Use maven wrapper

kumaraditya303:


yury-s:
Thanks for your interest in the project! Could you explain why we would need the wrapper?

kumaraditya303:
> Thanks for your interest in the project! Could you explain why we would need the wrapper?

@yury-s Maven is a better choice for maven as it is portable and cross platform where as installing maven on especially windows causes some PATH issues. Also because of this, no need to install maven on the machine and users building from source will have a better experience as they won't be required to install maven.

yury-s:
As far as I understand the wrapper may simplify Maven installation process but all our CI integrations work smoothly with the standard Maven installation and I also haven't heard any complains about using playwright-java with bare Maven in local setup. Considering that, I'm not sure we want this to be a part of the project at the current stage as it doesn't solve much while would require maintaining another flavor of dev setup. The functionality of the wrapper is gonna make it to Maven 3.7 so I'd just wait for that.

PR #91, opened 2020-12-06
WIP Added Dockerfile Fixes #70

### LINKED ISSUE #70: Dockerfile for playwright-java
Thanks again for creating this awesome project. 👏 

Do you guys happen to have a Dockerfile for playwright-java ? or any tips on how I can create one  ?

kumaraditya303:


Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM ubuntu:focal
```
yury-s, line 1: It should be easier to take `mcr.microsoft.com/playwright:focal` as base and install jdk and driver stuff on top of it.

kumaraditya303: The base image contains nodejs yarn and it is installing playwright from source which will make the image consume more bandwidth, because of this reason the python Docker Image is also built from source not by `mcr.microsoft.com/playwright:focal`

kumaraditya303:
@yury-s Docker test would pass after you merge #90

Dockerfile
yury-s: Sounds like the problem will be common for all language bindings, maybe we can build a base image with the bare minimum of deps required to launch browsers and then build an image with driver on top of it and use it for both python/java etc. I'll discuss this with other folks and come back.

yury-s: ok, we are going to split `mcr.microsoft.com/playwright:focal` into base which only includes browsers and their dependencies and move node.js specific stuff into an image derived from it. Java and Python bindings could be derived from the same base and install only language dependencies on top of it.

cc @aslushnikov

kumaraditya303: Marking as draft until the base image is created

kumaraditya303: PR on the way 🚀 https://github.com/microsoft/playwright/pull/4632

yury-s:
I've [added](https://github.com/microsoft/playwright-java/pull/97) a basic docker file that uses bare Maven without the wrapper, and is based on the current playwright image. We can iterate on that.

PR #117, opened 2020-12-13
docs: correct a document error and Fix an error about executing ./install_local_driver.sh

codeboyzhou:
1. Problem one
[Detail] docs(CONTRIBUTING.md): correct the filename of playwright-cli binary download script. It should be 'scripts/install_local_driver.sh' now instead of 'scripts/download_driver.sh'.

2. Problem two
[Detail] fix: an error 'No such file or directory' occurred when execute ./install_local_driver.sh. The detail error is './install_local_driver.sh: line 12: cd: ../driver/src/main/resources: No such file or directory', and its reason is that there is no 'resources' directory in the module 'driver', just create it and problem solved.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ cd playwright-java
 2. Run the following script to download playwright-cli binary for your platform into `driver/src/main/resources/driver/` directory. It will also install Playwright and download browser binaries for Chromium, Firefox and WebKit.
 
 ```bash
-scripts/download_driver.sh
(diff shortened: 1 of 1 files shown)


yury-s:
Sorry, this is a bit busted at the moment. I'll fix and update the documentation soon.

codeboyzhou:
> Sorry, this is a bit busted at the moment. I'll fix and update the documentation soon.

Never mind. Actually I have fixed both of them in this pull request. thank you for merging it.

PR #133, opened 2020-12-15
chore: drop api.json from repo, generate instead

yury-s:


.github/workflows/verify_api.yml
```diff
@@ -0,0 +1,46 @@
...
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Cache Downloaded Drivers
+        uses: actions/cache@v2
+        with:
+          path: driver-bundle/src/main/resources/driver
+          key: ${{ runner.os }}-drivers-${{ hashFiles('scripts/*') }}
+          restore-keys: ${{ runner.os }}-drivers
+      - name: Download drivers
+        shell: bash
```
mxschmitt, line 34: `shell: bash` is the default shell, I would remove it on all the steps since it probably won't be changed by the GHA team and is by that redundant.

mxschmitt (approved):

.github/workflows/verify_api.yml
yury-s: I remember it biting us on Windows, where it is not the default

yury-s: Removing for now.

mxschmitt: True 👍  I thought the verify workflow is enough to run on Linux only.

PR #154, opened 2020-12-19
chore: optimize driver download script for the bad network situation.

codeboyzhou:
Hi, I think maybe there is a minor issue that can be optimized.

When I try downloading the drivers by executing `scripts/download_driver_for_all_platforms.sh`, everything is ok here, but sometimes the download action could be aborted because of a bad network (for example `Connection timed out`). In this case, a driver may only download half of it, that is to say, the `zip` file is incomplete.

Then try it again, the download script will echo `Skipping driver download for...already exists` due to finding the driver zip file (incomplete), but in fact the download action is unsuccessful, and the half zip file can't be `unzip`. So we can't consider that the driver has been downloaded completely at this time, and one more download action also should not be skipped.

Although we can delete the incomplete zip file manually first and download again in this case, but I think we can consider whether those files we really want already exists (such as `playwright-cli`), if YES, that means it really downloaded successfully and completely. So we can really skip it at this time.

### DIFF
--- a/driver-bundle/pom.xml
+++ b/driver-bundle/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.microsoft.playwright</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>0.170.3</version>
+    <version>0.170.4-SNAPSHOT</version>
   </parent>
 
   <artifactId>driver-bundle</artifactId>
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.microsoft.playwright</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>0.170.3</version>
+    <version>0.170.4-SNAPSHOT</version>
   </parent>
 
   <artifactId>driver</artifactId>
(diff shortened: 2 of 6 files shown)


yury-s:
This doesn't seem to work yet:
image

Also if the download failed half-way it should be ok to just restart download with `-f` flag that will delete everything before downloading again. In general Playwright doesn't attempt to support network with poor connection where you cannot reliably download several hundred megabytes  that would require more effort when downloading browsers too. With that in mind and given that it is just a build script which is used locally and on CI I'd rather keep it simple and require starting over if anything went wrong.

codeboyzhou:
@yury-s Hi, thanks for your reply.

Here I have to say, sorry for that this pull request is closed unexpectedly because of my wrong operation.

Here I want to say, maybe I should check the CI result after pushing my code, and also I was only considering the platform `linux` and `macOS` in my code, what a shame on me!

I think about this question again, maybe your opinion is better, we should keep the script simple, especially it is only used locally. The code also seems a little difficult to understand if like what I pushed.

But there is one more thing to say, maybe we can put the `-f` flag and about its usage into the docs `CONTRIBUTING.md` to let other contributors know that we can use the `-f` flag if anything went wrong (actually I really forgot the `-f` flag before you told me, although I read the script code two days ago). what do you think of this advice? Of course, I'd like to help improve the docs if you don't mind.

yury-s:
The contributors instructions assume that the person is doing it first time, I'd rather keep it that way. We should probably add section for rolling new version of cli, another thing that might be useful is having help section in the script similar to https://github.com/microsoft/playwright/blob/master/browser_patches/export.sh#L18

PR #182, opened 2021-01-02
chore: bump maven dependency version to 0.172.3

codeboyzhou:
Bump maven dependency version to 0.172.3 in `README.md` and module `example/pom.xml`.


yury-s:
That version was published by mistake. Unfortunately there is no easy way to delete it from Maven Central.

codeboyzhou:
oh, I see.

PR #198, opened 2021-01-08
test: add new test 'Element Handle Select Text', Fixed webkit windows…

ephung01:
1. add new test - Test Element Handle Select Text scenarios
2. Fixed webkit windows platform check SSL Error message with regular expression in TestPageWaitForNavigation.shouldWorkWithClickingOnLinksWhichDoNotCommitNavigation()

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestElementHandleSelectText.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestElementHandleSelectText.java
@@ -1,0 +1,65 @@
+package com.microsoft.playwright;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class TestElementHandleSelectText extends TestBase {
+
+  @Test
+  void shouldSelectTextarea() {
+    page.navigate(server.PREFIX + "/input/textarea.html");
+    ElementHandle textarea = page.querySelector("textarea");
+    textarea.evaluate("textarea => textarea.value = 'some value'");
+    textarea.selectText();
+    if (isFirefox()) {
+      assertEquals(0, textarea.evaluate("el => el.selectionStart"));
+      assertEquals(10, textarea.evaluate("el => el.selectionEnd"));
+    } else {
+      assertEquals("some value", page.evaluate("() => window.getSelection().toString()"));
+    }
+  }
+
+  @Test
+  void shouldSelectInput() {
+    page.navigate(server.PREFIX + "/input/textarea.html");
+    ElementHandle input = page.querySelector("input");
+    input.evaluate("input => input.value = 'some value'");
+    input.selectText();
+    if (isFirefox()) {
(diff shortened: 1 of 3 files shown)


playwright/src/test/java/com/microsoft/playwright/TestPageWaitForNavigation.java
```diff
@@ -78,9 +77,8 @@ void shouldWorkWithClickingOnLinksWhichDoNotCommitNavigation() throws Interrupte
       fail("did not throw");
     } catch (PlaywrightException e) {
       // TODO: figure out why it is inconsistent on Linux WebKit.
-      assertTrue(e.getMessage().contains(expectedSSLError(browserType.name())) ||
-          (isWebKit() && getOS() == Utils.OS.LINUX && "Server required TLS certificate".equals(e.getMessage())),
-        "Unexpected exception: '" + e.getMessage() + "'");
+      System.out.println(e.getMessage());
```
yury-s, line 80: Revert this, you can pass it as second argument to assertTrue for debugging

playwright/src/test/java/com/microsoft/playwright/TestElementHandleSelectText.java
```diff
@@ -0,0 +1,91 @@
...
+  @Test
+  void shouldWaitForVisible() {
+    page.navigate(server.PREFIX + "/input/textarea.html");
+    ElementHandle textarea = page.querySelector("textarea");
+    textarea.evaluate("textarea => textarea.value = 'some value'");
+    textarea.evaluate("e => e.style.display = 'none'");
+
+    AtomicBoolean done = new AtomicBoolean(false);
+    // Async look for element timeout 3 seconds and return status if found or not
+    CompletableFuture<Boolean> completableFuture = CompletableFuture.supplyAsync(() -> {
```
yury-s, line 70: Playwright is not thread safe, it's designed to work on a single thread. I'll probably add a check that the APIs are called only from one thread.

These kinds of tests (should wait for ..) are essentially testing backend behavior. In typescript they are very easy to implement with promises but it's much more difficult with sync API. Given that there is a good coverage for the functionality upstream I was skipping such tests in java. Alternatively, if you really want to port such tests they can be converted to trigger the condition asynchronously from inside the page but in this case I suggest you drop this test in java (just leave an empty method similar to `shouldWaitForDisplayNoneToBeGone`)

ephung01: OK got it.  I will skip the last test case

playwright/src/test/java/com/microsoft/playwright/TestPageWaitForNavigation.java
ephung01: This test will failed on Webkit (Windows & Linux) due to different text in exception message.   Also in Mac OS webkit, I think issue probably due to multi-lines exception return so when check with contain text not found - should use regex to resolve it.

yury-s: What I meant was to revert debug printing line `System.out.println(e.getMessage());`, the reset is fine.

PR #201, opened 2021-01-10
fix(scripts): 'generate_api.sh' does not work on macOS

codeboyzhou:
Got a prompt message like this `playwright-java/driver-bundle/src/main/resources/driver/linux/node: cannot execute binary file` when I was executing `./scripts/generate_api.sh` on my `macOS Big Sur 11.1`, guess that is caused by different platform, so maybe we need to consider distinguishing the path of `playwright.sh` by platform.

### DIFF
--- a/scripts/generate_api.sh
+++ b/scripts/generate_api.sh
@@ -3,10 +3,19 @@
 set -e
 set +x
 
-trap "cd $(pwd -P)" EXIT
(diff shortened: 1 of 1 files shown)


yury-s:
Yeah, I don't think I've ever run it on any platform besides linux.

yury-s (approved):

PR #208, opened 2021-01-13
chore: automatically generate enum class for all simulatable devices

codeboyzhou:
Hi, I have an idea about when we use `playwright.devices().get("xxx")` method like this
`DeviceDescriptor pixel2 = playwright.devices().get("Pixel 2");`

As for some users of `playwright-java`, it's possible that they don't know how many and which devices that `playwright` can simulate (including me), in addition that the parameter of method `playwright.devices().get(String deviceName)` is `String`, there is no definite way to know about what value this parameter should be given.

So we can automatically generate the `enum` class (the class `Device`) according to `playwright.devices().keySet()` returned, still using `generate_api.sh`, combined with the class `DeviceEnumGenerator` (I add it in this commit), it's a little bit like class `ApiGenerator`.

If so, we can use `playwright.devices().get("xxx")` method like this
`DeviceDescriptor pixel2 = playwright.devices().get(Device.PIXEL_2.deviceName());`

Even if we don't know how many and which devices can be simulated, we can still use this method easily.


tools/api-generator/pom.xml
```diff
@@ -32,6 +32,11 @@
     </plugins>
   </build>
   <dependencies>
+    <dependency>
+      <groupId>com.microsoft.playwright</groupId>
```
yury-s, line 36: This introduces circular dependency as api-generator generates code for playwright but at the same time requires playwright. We should avoid that.

yury-s (commented):
In other languages (at least typescript, js, python, C#) we don't have named constant for the device names and just use string literal. It has never caused any issues so for the sake of simplicity and uniformity I'd keep it the same in java. If we'd like to see typed enum for the list of devices we should change it upstream too so that it's consistent across language bindings.

tools/api-generator/pom.xml
codeboyzhou: OK, I get it. Also that I didn't notice the circular dependency was introduced.
But is there any better way to know about which devices `playwright` can simulate? (except for debug the code)
:) Any reference docs in the upstream there?

yury-s: > But is there any better way to know about which devices `playwright` can simulate? (except for debug the code)
> :) Any reference docs in the upstream there?

https://playwright.dev/docs/api/class-playwright?_highlight=device#playwrightdevices is the upstream documentation. It simply points to the source file with the definitions of the devices. One could always print the list of keys in the map.

codeboyzhou: Thank you, It's clear and simple, now I'm closing this pull request.

PR #211, opened 2021-01-15
test: add new test 'Element Handle Type' and 'Element Handle Press'

ephung01:
add new test 'Element Handle Type' and 'Element Handle Press'


yury-s (approved):
I've updated string alignment in the code, otherwise looks great!

PR #235, opened 2021-01-29
fix: avoid NPE when saving storage state

jnizet:
When the path chosen to write the storage state is a relative path to the current directory (such as `Paths.get("state.json")`), the code tried to create directories for the parent path, which is null, thus causing a NullPointerException.
Moreover, the code also used a FileWriter, which uses the platform default encoding, which might not be valid for JSON, and which might not be able to encode some of the characters. The encoding is now set explicitly to UTF-8.


yury-s:
Thanks for the fix!

PR #238, opened 2021-01-29
feat: extend AutoCloseable in Playwright, BrowserContext, Browser and…

### LINKED ISSUE #237: [Feature] Consider implementing AutoCloseable
Playwright, BrowserContext, Browser and Page all have a `close()` method. And I guess that not calling them when we're done with those objects has a negative impact on resources.
To make that clearer, and more importantly, to be able to use standard Java constructs simplifying the proper handling of resources that must be closed, consider extening AutoCloseable in those 4 interfaces. It would allow, in Java to do:

```java
try (
    Playwright playwright = Playwright.create();
    Browser browser = playwright.chromium();
    BrowserContext browserContext = browser.newContext();
    Page page = browserContext.newPage()) {
  
}
```

and all those resources would be properly closed, even in case an exception is thrown, and without having to explicitly call `close()`.

Other JVM languages like Kotlin have similar constructs:

```kotlin
Playwright.create().use { playwright ->
    playwright.chromium().use { browser ->
        // ...
    }
}
```

But those constructs can't be used if those interfaces don't implement AutoCloseable. And code inspection tools also won't warn about missing close() calls.

jnizet:
… Page

fix #237


README.md
```diff
@@ -65,22 +65,21 @@ import java.util.List;
 
 public class PageScreenshot {
   public static void main(String[] args) throws Exception {
-    Playwright playwright = Playwright.create();
-    List<BrowserType> browserTypes = Arrays.asList(
+    try (Playwright playwright = Playwright.create()) {
```
yury-s, line 68: Let's also update [source files](https://github.com/microsoft/playwright-java/tree/master/examples/src/main/java/org/example) in the examples project.

README.md
```diff
@@ -124,22 +120,20 @@ import com.microsoft.playwright.*;
...
-    System.out.println(dimensions);
-    browser.close();
-    playwright.close();
+    try (Playwright playwright = Playwright.create();
+         Browser browser = playwright.firefox().launch(new BrowserType.LaunchOptions().withHeadless(false));
+         BrowserContext context = browser.newContext();
+         Page page = context.newPage()) {
+      page.navigate("https://www.example.com/");
+      Object dimensions = page.evaluate("() => {\n" +
+                                          "  return {\n" +
```
yury-s, line 129: nit: wrong alignment, let's keep it closer to the left for better readability of the readme file.

playwright/src/main/java/com/microsoft/playwright/Browser.java
```diff
@@ -25,7 +25,7 @@
  *
  * <p> A Browser is created via [{@code method: BrowserType.launch}]. An example of using a {@code Browser} to create a [Page]:
  */
-public interface Browser {
+public interface Browser extends AutoCloseable {
```
yury-s, line 28: These classes are generated, can you update [api-generator](https://github.com/microsoft/playwright-java/blob/master/tools/api-generator/src/main/java/com/microsoft/playwright/tools/ApiGenerator.java) with the new base interface? You can have a look at [`generate_api.sh`](https://github.com/microsoft/playwright-java/blob/master/scripts/generate_api.sh) to see how the generator is launched.  If the setup seems too tricky it's fine to commit it with the changes in the generated interfaces too and I can update the generator with the next roll of the playwright cli (I expect more changes there anyway).

playwright/src/test/java/com/microsoft/playwright/TestAutoClose.java
```diff
@@ -0,0 +1,18 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * This test simply makes sure that all 4 main interfaces implement AutoCloseable. If it compiles, then it passes.
+ */
+public class TestAutoClose {
+  @Test
+  void shouldAllowUsingTryWithResources() throws Exception {
+    try (Playwright playwright = Playwright.create();
+         Browser browser = playwright.chromium().launch();
```
yury-s, line 12: This will only run it for chromium, can you update the test similar to [`TestBrowser`](https://github.com/microsoft/playwright-java/blob/master/playwright/src/test/java/com/microsoft/playwright/TestBrowser.java) so that it launches the browser which is configure via `BROWSER` env?

playwright/src/test/java/com/microsoft/playwright/TestAutoClose.java
```diff
@@ -0,0 +1,18 @@
...
+         BrowserContext context = browser.newContext();
+         Page page = context.newPage()) {
+      page.navigate("about:config");
```
yury-s, line 15: I don't thing there is about:config in webkit, you could replace it with something more neutral like `assertEquals(2021, page.evaluate("() => 2021"));`

yury-s (commented):
Looks great, just a few comments.

jnizet:
Thanks for the review. I amended and rebased the PR.

yury-s (approved):

yury-s (approved):

PR #382, opened 2021-03-31
feat: roll driver, implement waitForUrl and new Video APIs

yury-s:


playwright/src/main/java/com/microsoft/playwright/impl/VideoImpl.java
```diff
@@ -16,27 +16,67 @@
...
+class VideoImpl extends LoggingSupport implements Video {
   private final PageImpl page;
-  private Path fullPath;
+  private final WaitableResult<ArtifactImpl> waitableArtifact = new WaitableResult<>();
+  private final boolean isRemote;
 
   VideoImpl(PageImpl page) {
     this.page = page;
+    BrowserImpl browser = page.context().browser();
+    isRemote = browser != null && browser.isRemote;
```
pavelfeldman, line 35: if (page.isClosed())
  waitableArtifact.complete(null);

pavelfeldman (approved):

playwright/src/main/java/com/microsoft/playwright/impl/VideoImpl.java
yury-s: This is unnecessary, waitForArtifact already does it.

PR #386, opened 2021-04-05
devops: adjust pushed Docker tags to be consistent with pw

mxschmitt:
The comment which I've removed was obsolete, we used it in the past but then decided to not use this CLI argument anymore.

See here for reference of the upstream format, with this PR its then in sync.

https://github.com/microsoft/playwright/blob/master/.github/workflows/publish_release.yml#L76

### DIFF
--- a/.github/workflows/publish_canary_docker.yml
+++ b/.github/workflows/publish_canary_docker.yml
@@ -7,7 +7,6 @@ on:
 jobs:
   publish-canary-docker:
     name: "publish to DockerHub"
-    # We use `docker push --all-tags` to push all tags which is a newly addition to docker
     runs-on: ubuntu-20.04
     if: github.repository == 'microsoft/playwright-java'
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_canary_docker.yml
```diff
@@ -23,5 +22,5 @@ jobs:
     - name: tag & publish
       run: |
         ./scripts/tag_image_and_push.sh playwright-java:localbuild-focal playwright.azurecr.io/public/playwright/java:next
-        ./scripts/tag_image_and_push.sh playwright-java:localbuild-focal playwright.azurecr.io/public/playwright/java:next-focal
+        ./scripts/tag_image_and_push.sh playwright-java:localbuild-focal playwright.azurecr.io/public/playwright/java:focal-next
```
yury-s, line 25: Revert this? According to https://mcr.microsoft.com/v2/playwright/tags/list we use *-<pw version>-<distro version> scheme, let's stick with that.

yury-s (changes requested):

.github/workflows/publish_canary_docker.yml
mxschmitt: you are right, this change was not correct. Good catch.

yury-s (approved):

Issue #418, opened 2021-04-28
[BUG] Playwright cannot download the browsers automatically


vidi42:
I'm trying to use Playwright in a Java EE environment with Zulu 11 and I always get this error messaged when launching chromium.
```
com.microsoft.playwright.PlaywrightException: Error {
  message='Failed to launch chromium because executable doesn't exist at C:\Users\Alex\AppData\Local\ms-playwright\chromium-857950\chrome-win\chrome.exe
Try re-installing playwright with "npm install playwright"
  name='Error
  stack='Error: Failed to launch chromium because executable doesn't exist at C:\Users\Alex\AppData\Local\ms-playwright\chromium-857950\chrome-win\chrome.exe
Try re-installing playwright with "npm install playwright"
    at Chromium._launchProcess (C:\Users\Alex\AppData\Local\Temp\playwright-java-12834975463618432611\package\lib\server\browserType.js:174:19)
    at async Chromium._innerLaunch (C:\Users\Alex\AppData\Local\Temp\playwright-java-12834975463618432611\package\lib\server\browserType.js:110:62)
    at async ProgressController.run (C:\Users\Alex\AppData\Local\Temp\playwright-java-12834975463618432611\package\lib\server\progress.js:79:28)
    at async Chromium.launch (C:\Users\Alex\AppData\Local\Temp\playwright-java-12834975463618432611\package\lib\server\browserType.js:78:25)
    at async BrowserTypeDispatcher.launch (C:\Users\Alex\AppData\Local\Temp\playwright-java-12834975463618432611\package\lib\dispatchers\browserTypeDispatcher.js:30:25)
    at async DispatcherConnection.dispatch (C:\Users\Alex\AppData\Local\Temp\playwright-java-12834975463618432611\package\lib\dispatchers\dispatcher.js:181:28)
}
    at com.microsoft.playwright.impl.WaitableResult.get(WaitableResult.java:54)
    at com.microsoft.playwright.impl.ChannelOwner.runUntil(ChannelOwner.java:88)
    at com.microsoft.playwright.impl.Connection.sendMessage(Connection.java:96)
    at com.microsoft.playwright.impl.ChannelOwner.sendMessage(ChannelOwner.java:79)
    at com.microsoft.playwright.impl.BrowserTypeImpl.launchImpl(BrowserTypeImpl.java:49)
    at com.microsoft.playwright.impl.BrowserTypeImpl.lambda$launch$0(BrowserTypeImpl.java:41)
[...]

PR #426, opened 2021-05-05
fix: manually pipe messages from child process sdtout/stderr

yury-s:
* Spawn a new dedicated thread to pump stderr/stdout from the driver child process to playwright process to mitigate possible problems with inheriting streams in Java Enterprise described in #418

* Close driver connection when reading from child process fails, this should help with  #347

### DIFF
--- a/driver-bundle/src/test/java/com/microsoft/playwright/TestInstall.java
+++ b/driver-bundle/src/test/java/com/microsoft/playwright/TestInstall.java
@@ -17,6 +17,7 @@
 package com.microsoft.playwright;
 
 import com.microsoft.playwright.impl.Driver;
+import com.microsoft.playwright.impl.StreamRedirectThread;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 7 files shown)


driver-bundle/src/test/java/com/microsoft/playwright/TestInstall.java
```diff
@@ -36,11 +37,13 @@ void playwrightCliInstalled() throws Exception {
       assertTrue(Files.exists(cli));
 
       ProcessBuilder pb = new ProcessBuilder(cli.toString(), "install");
-      pb.redirectError(ProcessBuilder.Redirect.INHERIT);
-      pb.redirectOutput(ProcessBuilder.Redirect.INHERIT);
       Process p = pb.start();
+      StreamRedirectThread stdoutThread = new StreamRedirectThread(p.getInputStream(), System.out);
```
danmana, line 41: Shouldn't this be `p.getOutpuStream()`, not input stream?

PR #449, opened 2021-05-20
fix: remote browser close in connect

yashLadha:
When the remote browser is closed, it only closes the transport
(WebSocket) and does not send the actual close message to the server
which happens in case of non-remote sessions.

This makes behavior similar to javascript client binding of playwright
and when done `browser.close`.

Fixes: #450

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/BrowserImpl.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/BrowserImpl.java
@@ -65,17 +65,11 @@ class BrowserImpl extends ChannelOwner implements Browser {
   }
 
   private void closeImpl() {
-    if (isConnectedOverWebSocket) {
-      try {
-        connection.close();
-      } catch (IOException e) {
-        throw new PlaywrightException("Failed to close browser connection", e);
-      }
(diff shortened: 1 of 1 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/BrowserImpl.java
```diff
@@ -65,17 +65,11 @@ public void close() {
...
-      try {
-        connection.close();
-      } catch (IOException e) {
-        throw new PlaywrightException("Failed to close browser connection", e);
-      }
-      notifyRemoteClosed();
-      return;
-    }
     try {
       sendMessage("close");
```
yury-s, line 69: When we connected over web socket it's ok to simply close the connection, the server should handle that gracefully anyway (e.g. when the network connection is aborted). If this change fixes a particular bug please add a test (see [TestBrowserTypeConnect.java](https://github.com/microsoft/playwright-java/blob/master/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java) for examples).

yury-s (changes requested):

yury-s:
Closing for now.

PR #519, opened 2021-07-17
feat: provide jbang-catalog to run playwright cli

maxandersen:
current docs explains how when you have a maven project you can run the cli.

with this PR you can just use jbang.

if you don't have jbang installed already see jbang.dev/downloads or just run this first:

```
curl -Ls https://sh.jbang.dev | bash -s - app setup
```

then you can do:
```
jbang playwright@microsoft/playwright-java codegen -o Example.java
```

for now you can try it with the following which runs it from my branch:
```
jbang playwright@maxandersen/playwright-java/jbang codegen -o Example.java
```

I also opened https://github.com/microsoft/playwright/pull/7696 which adds small header to generated java code 
that enables you to run the generated scripts directly. if that happens you can just do `jbang Example.java` 

Thank you

### DIFF
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -1,0 +1,10 @@
+{
+  "catalogs": {},
+  "aliases": {
+    "playwright": {
+      "script-ref": "scripts/playwright.java",
+      "description": "Playwright lets you automate Chromium, Firefox and Webkit with a single API. \nWith this cli you can install, trace, generate pdf and screenshots and more.\nExample on how to record and run a script:\n```\n  jbang playwright@microsoft/playwright-java codegen -o Example.java`\n  jbang --deps com.microsoft.playwright:playwright:RELEASE Example.java\n```"
+    }
+  },
+  "templates": {}
+}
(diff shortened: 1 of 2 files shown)


yury-s:
I'm getting the following error while trying to run the command from the description:

```sh
$ jbang playwright@maxandersen/playwright-java/jbang codegen -o Example.java
[jbang] Resolving dependencies...
[jbang]     Resolving com.microsoft.playwright:playwright:RELEASE...[jbang] [ERROR] Could not resolve dependency com.microsoft.playwright:playwright:RELEASE
```

I wonder if it has something to do with multiple versions of playwright cached on my laptop.

yury-s:
> I wonder if it has something to do with multiple versions of playwright cached on my laptop.

Yeah, looks so, deleting `~/.m2/repository/com/microsoft/playwright` resolved the issue.

yury-s (approved):

maxandersen:
Interesting - haven't seen that error. It's a standard maven syntax so should work. Would have been great to have known what you had in your folder before deleting :)

We can also set the exact version in the catalog - but you'll then need to update it on release.

yury-s:
Let's keep it as is, if anyone else sees the same issue we can revisit it. Updating hard-coded version is worse IMO.

PR #533, opened 2021-07-29
fix: set apiName for method calls in metadata

yury-s:


playwright/src/main/java/com/microsoft/playwright/impl/Connection.java
```diff
@@ -94,6 +95,12 @@ public String toString() {
     }
   }
 
+  String setApiName(String name) {
+    String previous = apiName;
+    apiName = name;
+    return previous;
```
mxschmitt, line 101: whats the reason of returning the previous one here?

playwright/src/main/java/com/microsoft/playwright/impl/ChannelOwner.java
```diff
@@ -69,9 +69,18 @@ void disconnect() {
...
+    return withLogging(apiName, new WaitForEventLogger<>(this, apiName, code));
   }
 
+  @Override
+  <T> T withLogging(String apiName, Supplier<T> code) {
+    String previousApiName = connection.setApiName(apiName);
+    try {
+      return super.withLogging(apiName, code);
+    } finally {
+      connection.setApiName(previousApiName);
```
mxschmitt, line 81: having it on the connection feels somehow wrong and racy. But not familiar with pw-Java.

mxschmitt (approved):

playwright/src/main/java/com/microsoft/playwright/impl/ChannelOwner.java
yury-s: Connection is required to be used with exclusive access and hence it guarantees here that we'll not have a race. I agree thought that Connection is somewhat abused for this purpose, we can refactor this if it becomes an issue.

PR #583, opened 2021-08-24
feat(roll): roll Playwright to 1.15.0-next-1629487941000

mxschmitt:
Applies:
- https://github.com/microsoft/playwright/pull/8019
- https://github.com/microsoft/playwright/pull/8329


scripts/CLI_VERSION
```diff
@@ -1 +1 @@
-1.15.0-next-1629322356000
+1.15.0-next-1629487941000
```
mxschmitt, line 1: note: this is the commit after the protocol changes, not ToT to reduce the scope of this PR.

playwright/src/main/java/com/microsoft/playwright/impl/Connection.java
```diff
@@ -177,11 +180,8 @@ private JsonArray currentStackTrace() {
     return result;
   }
 
-  public ChannelOwner waitForObjectWithKnownName(String guid) {
-    while (!objects.containsKey(guid)) {
-      processOneMessage();
-    }
-    return objects.get(guid);
+  public PlaywrightImpl initializePlaywright() {
```
yury-s, line 183: It's a layering violation when the connection knows about playwright impl types but I think it's fine for now.

yury-s (approved):

PR #586, opened 2021-08-28
chore: Added user agent as default headers in browser type

Archish27:


playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
```diff
@@ -73,6 +73,7 @@ private Browser connectImpl(String wsEndpoint, ConnectOptions options) {
           slowMo = Duration.ofMillis(options.slowMo.intValue());
         }
       }
+      headers.put("User-Agent", getUserAgent());
```
yury-s, line 76: User-agent passed in the headers parameter should override default value. So this statement should probably moved up.

yury-s (commented):
Looks good, please add a test for this.

playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
Archish27: fixed

Archish27: Have added tests.

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
```diff
@@ -169,6 +169,17 @@ void shouldSendExtraHeadersWithConnectRequest() throws Exception {
...
   }
 
+  @Test
+  void shouldSetDefaultHeadersAsUserAgentWithConnectRequest() throws Exception {
+    try (WebSocketServerImpl webSocketServer = WebSocketServerImpl.create()) {
+      try {
+        browserType.connect("ws://localhost:" + webSocketServer.getPort() + "/ws");
+      } catch (Exception e) {
+      }
+      assertTrue("Playwright".contains(webSocketServer.lastClientHandshake.getFieldValue("User-Agent")));
```
yury-s, line 179: I believe it should be the other way around: `assertTrue(fieldValue.contains("Playwright"));`

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
```diff
@@ -83,7 +83,7 @@ private static BrowserServer launchBrowserServer(BrowserType browserType) {
 
   @Override
   @BeforeAll
-  // Hide base class method to launch browser server and connect to it.
+    // Hide base class method to launch browser server and connect to it.
```
yury-s, line 86: revert this

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
```diff
@@ -253,7 +264,8 @@ void shouldThrowWhenCallingWaitForNavigationAfterDisconnect() throws Interrupted
     }
     assertFalse(browser.isConnected());
     try {
-      page.waitForNavigation(() -> {});
+      page.waitForNavigation(() -> {
```
yury-s, line 267: revert?

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
```diff
@@ -265,7 +277,8 @@ void shouldRejectNavigationWhenBrowserCloses() {
     Browser remote = browserType.connect(wsEndpoint);
     Page page = remote.newPage();
 
-    server.setRoute("/one-style.css", r -> {});
+    server.setRoute("/one-style.css", r -> {
```
yury-s, line 280: revert?

playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
```diff
@@ -60,7 +61,8 @@ public Browser connect(String wsEndpoint, ConnectOptions options) {
   private Browser connectImpl(String wsEndpoint, ConnectOptions options) {
     try {
       Duration timeout = Duration.ofDays(1);
-      Map<String, String> headers = Collections.emptyMap();
+      Map<String, String> headers = new HashMap<>();
+      headers.put("User-Agent", getUserAgent());
```
yury-s, line 65: I see two problems with this:
1. This way the user-agent field is going to be always overridden (at line are 72) if any headers are passed in the options. We want it to be overridden only if 'User-Agent' header is present in the options.
2. The header name should be matched case-insensitively, i.e. 'user-agent'  header option should still override default value.

playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
```diff
@@ -105,6 +107,14 @@ private Browser connectImpl(String wsEndpoint, ConnectOptions options) {
     }
   }
 
+  private String getUserAgent() {
+    return String.format("Playwright/%s (%s/%s/%s)",
+      getClass().getPackage().getImplementationVersion(),
```
yury-s, line 112: This is java package version which can have a form of `1.15.0-SNAPSHOT` and even the release version is slightly different from the driver version. Do you want to see here driver version or playwright java version?

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
Archish27: I have formatted using IntelliJ hence the above changes

playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
Archish27: Need to get playwright version i.e `1.15.0`

playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
Archish27: Updated

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
Archish27: ahh silly mistake fixed this.

playwright/src/main/java/com/microsoft/playwright/impl/BrowserTypeImpl.java
yury-s: Then this will break in some cases. Also it returns `null` on the to setups that I tried, is it expected?

yury-s (commented):
When this [upstream PR](https://github.com/microsoft/playwright/pull/8566) lands we'll get user agent from the driver for free, so we can just wait for that.

yury-s:
After the latest driver roll this should already work out of the box as we now share the same connect logic with nodejs.

Issue #610, opened 2021-09-17
[Bug]: shouldSetBodySizeTo0WhenThereWasNoResponseBody:81 expected: <0> but was: <5>


yury-s:
### Playwright version

1.15.0-SNAPSHOT

### What happened? / Describe the bug

After recent roll chromium is failing across all platforms:

TestPageNetworkSizes.shouldSetBodySizeTo0WhenThereWasNoResponseBody:81 expected: <0> but was: <5>

Channel response:
```
2021-09-17T21:54:54.523Z ◀ RECV {"id":8,"result":{"sizes":{"requestBodySize":0,"requestHeadersSize":502,"responseBodySize":5,"responseHeadersSize":109}}}
```

PR #616, opened 2021-09-20
test: fix sizes test by using no chunked requests

mxschmitt:
Fixes https://github.com/microsoft/playwright-java/issues/610

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestPageNetworkSizes.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestPageNetworkSizes.java
@@ -53,13 +53,12 @@ public class TestPageNetworkSizes extends TestBase {
   }
 
   @Test
-  @Disabled("responseBodySize == 16")
   void shouldSetBodySizeHeadersSizeAndTransferSize() throws ExecutionException, InterruptedException {
     server.setRoute("/get", exchange -> {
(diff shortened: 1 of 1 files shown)


playwright/src/test/java/com/microsoft/playwright/TestPageNetworkSizes.java
```diff
@@ -53,13 +53,12 @@ void shouldSetBodySizeTo0IfThereWasNoBody() {
...
 
   @Test
-  @Disabled("responseBodySize == 16")
   void shouldSetBodySizeHeadersSizeAndTransferSize() throws ExecutionException, InterruptedException {
     server.setRoute("/get", exchange -> {
       // In Firefox, |fetch| will be hanging until it receives |Content-Type| header
       // from server.
       exchange.getResponseHeaders().add("Content-Type", "text/plain; charset=utf-8");
-      exchange.sendResponseHeaders(200, 0);
+      exchange.sendResponseHeaders(200, 6);
```
mxschmitt, line 61: there seems unfortunately no better way to do this.

yury-s: "If the response length parameter is {@code zero}, then chunked transfer encoding is used..." - yeah, this is unfortunate

yury-s (approved):

PR #699, opened 2021-11-13
devops: use main branch instead of master

mxschmitt:


.github/workflows/verify_api.yml
```diff
@@ -2,23 +2,21 @@ name: Verify API
...
-      - master
+      - main
       - release-*
     paths:
       - 'scripts/**'
       - 'api-generator/**'
 jobs:
   verify:
     timeout-minutes: 30
-    strategy:
```
mxschmitt, line 20: drive-by: only relevant if a matrix gets used. In this case there isn't used any.

PR #722, opened 2021-11-27
fix(scripts): install_local_driver.sh can't download driver successfully

codeboyzhou:
Hi, I got a problem when try to download the driver by `install_local_driver.sh`, I found that the download url is incorrect, and it resulted in download failure finally, so I try to read its code and fixed this issue, referred to the `download_driver_for_all_platforms.sh`. It works now, I have a test on my macOS machine.

### DIFF
--- a/driver-bundle/src/main/java/com/microsoft/playwright/impl/DriverJar.java
+++ b/driver-bundle/src/main/java/com/microsoft/playwright/impl/DriverJar.java
@@ -28,7 +28,7 @@ public class DriverJar extends Driver {
   private static final String PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD = "PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD";
   private final Path driverTempDir;
 
-  DriverJar() throws IOException, URISyntaxException, InterruptedException {
+  DriverJar() throws IOException {
     driverTempDir = Files.createTempDirectory("playwright-java-");
     driverTempDir.toFile().deleteOnExit();
   }
--- a/driver-bundle/src/test/java/com/microsoft/playwright/TestInstall.java
+++ b/driver-bundle/src/test/java/com/microsoft/playwright/TestInstall.java
@@ -24,7 +24,6 @@ import java.nio.file.Path;
 import java.util.Collections;
 import java.util.concurrent.TimeUnit;
 
-import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class TestInstall {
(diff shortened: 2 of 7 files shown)


codeboyzhou:
Oops, maybe I should say sorry for this pull request, I have produced two commits on it, the commit [c9cbb67](https://github.com/microsoft/playwright-java/pull/722/commits/c9cbb67839d63a7027bb07a3748d48f234cffdb3) should be a new one, but I forgot to switch my git branch.

Emm...but the second commit is very simple and just only considering accept some suggestions from IntelliJ IDEA to avoid code warnings, maybe you can also review it by the way :)

yury-s:
> Hi, I got a problem when try to download the driver by `install_local_driver.sh`

The fix looks good. The reason the file rotted is we don't use it anywhere in our dev process or CI/CD, so I'm tempted to drop it altogether. Do you find it useful and have a use case that would justify keeping it in the  repo?

Meanwhile landing the fix.


> Oops, maybe I should say sorry for this pull request, I have produced two commits on it, the commit [c9cbb67](https://github.com/microsoft/playwright-java/pull/722/commits/c9cbb67839d63a7027bb07a3748d48f234cffdb3) should be a new one, but I forgot to switch my git branch.
> 
> Emm...but the second commit is very simple and just only considering accept some suggestions from IntelliJ IDEA to avoid code warnings, maybe you can also review it by the way :)

It's usually better to push such things separately but all of them look safe so let's land them to avoid double work on your end.

yury-s (approved):

codeboyzhou:
> Do you find it useful and have a use case that would justify keeping it in the repo?

No, I did not. Actually I also wanted to drop the file `install_local_driver.sh`, maybe we don't need it anymore.

In addition, I have an another idea, perhaps we can add an option for `download_driver_for_all_platforms.sh`, for instance `./download_driver_for_all_platforms.sh --mac`, to download the driver only for macOS platform, it can be `--linux` or `--win32`, if there is no any options, we download drivers for all of them.

What do you think of this idea?

codeboyzhou:
> It's usually better to push such things separately but all of them look safe so let's land them to avoid double work on your end.

Thanks, I will avoid this next time.

yury-s:
> In addition, I have an another idea, perhaps we can add an option for `download_driver_for_all_platforms.sh`, for instance `./download_driver_for_all_platforms.sh --mac`, to download the driver only for macOS platform, it can be `--linux` or `--win32`, if there is no any options, we download drivers for all of them.
> 
> What do you think of this idea?

I don't think we have a use case for that.

codeboyzhou:
> I don't think we have a use case for that.

ok, I got you :)

PR #723, opened 2021-11-29
feat(roll): roll Playwright 1.17.0-beta-1638222602000

mxschmitt:


scripts/CLI_VERSION
```diff
@@ -1 +1 @@
-1.17.0-rc1
+1.17.0-beta-1638222602000
```
yury-s, line 1: This transition looks odd, I'd much rather have a reference to mere `1.17.0` which corresponds to the node.js release

yury-s (approved):

PR #770, opened 2022-01-13
chore: set java runtime version

aslushnikov:
This is used to properly set Playwright user agent for all the
requests it does.


yury-s (commented):
Let's add a test for this?

yury-s:
Subsumed into https://github.com/microsoft/playwright-java/pull/774

PR #774, opened 2022-01-15
fix: pass required env variables for new driver

### LINKED ISSUE #772: internal: setup env variables when launching driver
1.18 driver assumes the following environment variables are set:

```
PW_CLI_TARGET_LANG=java
PW_CLI_TARGET_LANG_VERSION=<version>
PW_CLI_DISPLAY_VERSION=CLI_VERSION (of the language)
```

version should not be too granular to avoid noise; IMO major seems to be good for Java.

aslushnikov:
Fixes #772


playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
```diff
@@ -237,4 +237,23 @@ static String toJsRegexFlags(Pattern pattern) {
     }
     return regexFlags;
   }
+
+  static void setDriverEnvironmentVariables(ProcessBuilder pb) {
```
yury-s, line 241: I'd rather we moved it to Driver.setDriverEnvironmentVariables. `PW_CLI_DISPLAY_VERSION` should also be set in the same method I believe.

playwright/src/test/java/com/microsoft/playwright/TestGlobalFetch.java
```diff
@@ -16,6 +16,16 @@
...
 
 public class TestGlobalFetch extends TestBase {
+  @Test
+  void shouldHaveJavaInDefaultUesrAgent() throws ExecutionException, InterruptedException {
+    APIRequestContext request = playwright.request().newContext(new APIRequest.NewContextOptions());
+    Future<Server.Request> serverRequest = server.futureRequest("/empty.html");
+    APIResponse response = request.get(server.EMPTY_PAGE);
+    assertTrue(response.ok());
+    assertEquals(server.EMPTY_PAGE, response.url());
+    assertTrue(serverRequest.get().headers.get("user-agent").get(0).contains("java"));
```
yury-s, line 26: Let's also check version?

driver/src/main/java/com/microsoft/playwright/impl/Driver.java
```diff
@@ -65,6 +65,25 @@ public Path driverPath() {
     return driverDir().resolve(cliFileName);
   }
 
+  public static void setRequiredEnvironmentVariables(ProcessBuilder pb) {
+    if (!pb.environment().containsKey("PW_CLI_TARGET_LANG")) {
+      pb.environment().put("PW_CLI_TARGET_LANG", "java");
```
yury-s, line 70: looks like PW_CLI_TARGET_LANG /PW_CLI_TARGET_LANG_VERSION are used beyond just codegen, let's rename  them (in separate pr) to PW_LANG / PW_LANG_VERSION before it's too late?

yury-s (approved):

driver/src/main/java/com/microsoft/playwright/impl/Driver.java
aslushnikov: Let's do this in a follow-up and on tip-of-tree only.

I filed an issue for myself regarding this: https://github.com/microsoft/playwright/issues/11450

playwright/src/test/java/com/microsoft/playwright/TestGlobalFetch.java
aslushnikov: done

PR #790, opened 2022-01-24
fix: CLI don't download browers automatically and set env accordingly

mxschmitt:


driver-bundle/src/main/java/com/microsoft/playwright/impl/DriverJar.java
```diff
@@ -40,9 +40,10 @@ public DriverJar() throws IOException {
   }
 
   @Override
-  protected void initialize(Map<String, String> env) throws Exception {
+  protected void initialize(Map<String, String> env, Boolean installBrowsers) throws Exception {
```
yury-s, line 43: nit: we could make `installBrowsers` a method on the base `Driver` and avoid this parameter

yury-s (approved):

PR #837, opened 2022-03-08
devops: publish no arm Docker for now

mxschmitt:


utils/docker/publish_docker.sh
```diff
@@ -113,6 +113,5 @@ publish_docker_manifest () {
   done
 }
 
-publish_docker_images_with_arch_suffix focal amd64
 publish_docker_images_with_arch_suffix focal arm64
```
aslushnikov, line 116: shall we drop this as well?

aslushnikov (approved):

utils/docker/publish_docker.sh
mxschmitt: had a typo, arm should have been deleted.

Issue #842, opened 2022-03-11
[Feature] LocatorAssertions.setDefaultTimeout()


N1ro:
Hello everyone,

I'd like to request a change for `LocatorAssertions.setDefaultTimeout() `

Why?
Currently, assertions are performed like below with `setTimeout`, we'd like to know if its possilbe to add `LocatorAssertions.setDefaultTimeout()` so we can setup locationAssertions timeout globallly as its possible with `BrowserContext` and `Page` classes, otherwise we'd have to do it individually like so in the below example.

`assertThat(page.locator(  <some locator> ))
            .hasText( expectedMessage, new LocatorAssertions.HasTextOptions().setTimeout( 30000 ) );`
            

Please let us know if you need more info.
           
 Thanks
 Niro

chaganiu:
I just ran into this myself.  one of our apps is slow and the page/browsercontext default timeouts work fine but fail with locator assertions because they take longer than 5 seconds.  This causes us to either write `LocatorAssertions.HasTextOptions().setTimeout( 30000 )` over and over or we have to add

`locator.waitFor()` before asserting.

PR #855, opened 2022-03-16
Default Timeout for Locator Assertions

uchagani:
I took a shot at implementing #842.  I'm not sure if this is the proper approach but I'm hoping it can at least get the ball rolling.  Happy to make changes with any guidance you can provide.

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestLocatorAssertions.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestLocatorAssertions.java
@@ -176,7 +176,7 @@ public class TestLocatorAssertions extends TestBase {
 
   @Test
   void defaultTimeoutFail() {
-    LocatorAssertions.TimeoutOptions.setDefaultTimeout(0);
+    LocatorAssertions.TimeoutOptions.setDefaultTimeout(1000);
     page.setContent("<div>Text 1</div><div>Text 3</div>");
     Locator locator = page.locator("div");
     page.evaluate("setTimeout(() => {\n" +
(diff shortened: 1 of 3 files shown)


uchagani:
@yury-s @mxschmitt I'm not sure why some of these tests are failing.  Doesn't seem to have anything to do with this PR.  I tried to kick off a re-run of the jobs but I'm guessing permissions issues don't allow that.  Would you be able to kick off a re-run to see if these are flaky tests?

uchagani:
I found some bugs i'm working through.  closing for now.

uchagani:
@mxschmitt it seems the node version has a way to configure timeouts via a config: https://playwright.dev/docs/test-timeouts#set-expect-timeout-in-the-config.  If my understanding is correct, this would be similar to a global LocatorAssertions timeout, correct?

mxschmitt:
Yeah but it affects Python and .NET too. Please file an issue in https://github.com/microsoft/playwright

uchagani:
ah ok, makes sense.

playwright/src/test/java/com/microsoft/playwright/TestLocatorAssertions.java
```diff
@@ -165,6 +165,66 @@ void hasTextWTextArrayFailOnNotEmpty() {
...
+    page.evaluate("setTimeout(() => {\n" +
+      "  div.innerHTML = \"<p>Text 1</p><p>Text 2</p>\";\n" +
+      "}, 100);");
+    try {
+      // Should normalize whitespace.
+      assertThat(locator).hasText(new String[] {"Text  1", "Text   2"});
+      fail("did not throw");
+    } catch (AssertionFailedError e) {
+      assertEquals("[Text  1, Text   2]", e.getExpected().getStringRepresentation());
+      assertEquals("null", e.getActual().getStringRepresentation());
```
uchagani, line 190: @yury-s @mxschmitt I could use some help here.  `e.getExpected()`returns null when the default timeout is set to 0.  Any value great than 0 will work as expected but i'm not sure why that is.

uchagani:
Re-opening this PR as @pavelfeldman said it was okay to do this in just Java.  I could use some help though (see comment above).

playwright/src/test/java/com/microsoft/playwright/TestLocatorAssertions.java
uchagani: `getExpected()` and `getActual()` both return null:

image

uchagani:
I'll start working on the upstream part this weekend

PR #866, opened 2022-03-25
fix(tracing): support explicit StartOptions().setSources(false)

leonard84:
Prior to this commit using `new Tracing.StartOptions().setSources(false) was
was wrongfully treated as `.setSources(true)` since it only tested the `Boolean` to be non-null.
Also, `includeSources` was converted into a local variable as it is not used anywhere else.


yury-s (approved):

yury-s:
The patch looks good! Please sign the CLA above so that we could merge your PR.

leonard84:
@yury-s done, had to get legal to give the green light first.

PR #879, opened 2022-03-31
fix(driver): set driver instance to null if an exception is thrown du…

### LINKED ISSUE #878: [BUG] Calling Playwright.create a second time with bad env results in a PlaywrightImpl object instead of Exception again
**Context:**
- Playwright Version: [what Playwright version do you use?] 1.20
- Operating System: [e.g. Windows, Linux or Mac] Mac
- Browser: [e.g. All, Chromium, Firefox, WebKit] All
- Extra: [any specific details about your environment]

**Code Snippet**

```java
@Test
public void testPlaywrightCreate() {
    Playwright.CreateOptions options = new Playwright.CreateOptions().setEnv(Map.of("PLAYWRIGHT_BROWSERS_PATH", "/some/bad/path"));
    
    // First time throws as expected.
    Assertions.assertThrows(RuntimeException.class, () -> Playwright.create(options));

    // Second time does NOT throw but returns `PlaywrightImpl`
    Assertions.assertThrows(RuntimeException.class, () -> Playwright.create(options));
}
```

**Describe the bug**

I'm writing an extension that manages the lifecycle of Playwright for the user.  In this situation there may be calls to `Playwright.create` multiple times.  The situation is:

1.  Before test start create Playwright with bad options
2. Playwright throws `Failed to install browsers Error: ENOENT: no such file or directory, mkdir '/some'` as expected
3. After test attempt to close playwright if created.  At this point the same code that is used to create playwright is called again.  The assumption is that it will throw again but it doesn't.  A `PlaywrightImpl` object is returned instead.

uchagani:
…ring creation or initialization.  Fixes #878

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,14 +31,18 @@ jobs:
     - name: Build & Install
       run: mvn -B install -D skipTests --no-transfer-progress
     - name: Run tests
-      run: mvn test --no-transfer-progress --fail-at-end
+      run: mvn test -DexcludedGroups=isolated --no-transfer-progress --fail-at-end
       env:
         BROWSER: ${{ matrix.browser }}
     - name: Run tracing tests w/ sources
-      run: mvn test --no-transfer-progress --fail-at-end -D test=*TestTracing*
+      run: mvn test -DexcludedGroups=isolated --no-transfer-progress --fail-at-end -D test=*TestTracing*
       env:
         BROWSER: ${{ matrix.browser }}
         PLAYWRIGHT_JAVA_SRC: src/test/java
+    - name: Run driver throw tests
+      run: mvn test -Dgroups=driverThrowTest --no-transfer-progress --fail-at-end
+      env:
+        BROWSER: ${{ matrix.browser }}
     - name: Test Spring Boot Starter
       shell: bash
       env:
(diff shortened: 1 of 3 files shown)


uchagani:
hmm not sure why it's failing on windows.  the tests failing on ubuntu seem unrelated (`TestWheel.shouldWorkWhenTheEventIsCanceled`)

yury-s (approved):

yury-s (changes requested):
This error is quite consistent across runners, let's fix it before landing:

```
Error:  Failures: 
Error:    TestInstall.playwrightDriverAlternativeImpl:[86](https://github.com/microsoft/playwright-java/runs/5775734111?check_suite_focus=true#step:7:86) Expected java.lang.RuntimeException to be thrown, but nothing was thrown.
Error:    TestInstall.shouldThrowWhenBrowserPathIsInvalid:53 Expected java.lang.RuntimeException to be thrown, but nothing was thrown.
```

I believe the new test should run in a new VM process as otherwise `Driver.instance` may be initialized by other tests.

yury-s:
> (`TestWheel.shouldWorkWhenTheEventIsCanceled`)

This one has been flaky, don't worry about it, it needs to be addressed separately.

uchagani:
@yury-s  I added the test to run on it's own.

yury-s (approved):

Issue #836, opened 2022-03-08
ARM based Docker image


SodaDev:
Could you provide arm based docker images?

vcrego:
We're really looking forward for this change to be included asap, we have got more than hundreds clients are awaiting to go for arm, once this change is implemented then we should be able to move ahead. Request you to please push this asap. Thanks.

steveseaney:
This change is critical for client base as well as the broader audience of applications using the AWS ARM stack.

omar-luna-rego:
This change will be a massive improvemnt in the AWS ARM stack an a great help in our AWS architecture

PR #883, opened 2022-04-06
Add Linux/arm64 support

otterley:
Allow playwright-java to use existing support in https://github.com/microsoft/playwright for Linux on arm64, available since version 1.17.

Since playwright itself only supports Linux/arm64 on Ubuntu today, so must playwright-java until support for other distros is added.

All tests currently pass on this distribution and architecture.

References: #836

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ Install git, Java JDK (version >= 8), Maven (tested with version 3.6.3), on Ubun
 just run the following command:
 
 ```sh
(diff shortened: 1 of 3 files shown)


yury-s:
> Since playwright itself only supports Linux/arm64 on Ubuntu today, so must playwright-java until support for other distros is added.

We actually run on macOS arm too (m1), would be nice to add it too (we could do it in a separate PR).

yury-s (approved):
We need to add a bot to run at least trivial tests for arm.

PR #890, opened 2022-04-11
chore: add arm64 Docker image

mxschmitt:
Fixes https://github.com/microsoft/playwright-java/issues/836


yury-s (approved):
Let's setup a test for the arm image.

yury-s:
Note that I'm reverting Arm driver changes for now: https://github.com/microsoft/playwright-java/pull/892 so we should wait before landing this PR.

utils/docker/Dockerfile.focal
```diff
@@ -13,7 +13,8 @@ RUN apt-get update && \
     # Create the pwuser
     adduser pwuser
 
-ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
+ARG PW_TARGET_ARCH
```
yury-s, line 16: We should be able to use `uname -m` to detect target architecture.

yury-s (approved):

utils/docker/Dockerfile.focal
mxschmitt: We can't set env vars dynamically with a Dockerfile so I saw no other way around it. :/ 
Or wdyt?

yury-s: Yeah, that's a bummer, I guess the best workaround at the moment is this:
```
RUN export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-$(dpkg --print-architecture)
```
Or even writing this env variable into .bashrc but I agree this is somewhat cumbersome.

yury-s (approved):

utils/docker/Dockerfile.focal
yury-s: Yeah, that's a bummer, I guess the best workaround at the moment is this:
```
RUN export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-$(dpkg --print-architecture)
```
Or even writing this env variable into .bashrc but I agree this is somewhat cumbersome.

mxschmitt: > Yeah, that's a bummer, I guess the best workaround at the moment is this:
> 
> ```
> RUN export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-$(dpkg --print-architecture)
> ```
> 
> Or even writing this env variable into .bashrc but I agree this is somewhat cumbersome.

This won't work, since the export only sets it for the current active bash/sh session which does not re-used / envs get pulled out of it.

> Or even writing this env variable into .bashrc but I agree this is somewhat cumbersome.

Imagine a user does put "java -jar ..." as a Docker entrypoint, it would never invoke bash/sh so the custom env var does not get loaded / injected.

---

So I think having it with a custom arg might be the best solution for now.

PR #892, opened 2022-04-12
Revert "Add Linux/arm64 support (#883)"

yury-s:
This reverts commit 7eddd2d2b2ee1230381c095ebdaa1148218aa15d.

Reverting this change as we are going to branch for the next release and amr support
* is not covert by tests yet
* inclreases the size of the driver bundle which we'll likely want to fix before
release.

References https://github.com/microsoft/playwright-java/issues/836

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ Install git, Java JDK (version >= 8), Maven (tested with version 3.6.3), on Ubun
 just run the following command:
 
 ```sh
-sudo apt-get install git openjdk-11-jdk maven unzip
+sudo apt-get install git openjdk-11-jdk maven
 ```
 
 ### Getting the Code
--- a/driver-bundle/src/main/java/com/microsoft/playwright/impl/DriverJar.java
+++ b/driver-bundle/src/main/java/com/microsoft/playwright/impl/DriverJar.java
@@ -139,17 +139,11 @@ public class DriverJar extends Driver {
 
   private static String platformDir() {
     String name = System.getProperty("os.name").toLowerCase();
-    String arch = System.getProperty("os.arch").toLowerCase();
-
     if (name.contains("windows")) {
       return "win32_x64";
     }
     if (name.contains("linux")) {
-      if (arch.equals("aarch64")) {
-        return "linux-arm64";
-      } else {
-        return "linux";
-      }
+      return "linux";
     }
     if (name.contains("mac os x")) {
       return "mac";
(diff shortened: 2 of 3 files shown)


aslushnikov (approved):

otterley:
Which release are you targeting Linux/ARM support to be available?

otterley:
Which release are you targeting Linux/ARM support to be available? I have at least one customer who is depending on this support -- which is already available in the Node.js version of Puppeteer -- to be available soon. 

Can you speak more about the negative impact of including the support? It's not super clear from this PR. Any data you can cite would be helpful.

otterley:
Which release are you targeting Linux/ARM support to be available? 

Can you speak more about the negative impact of including the support that led to this reversion? It's not super clear from this PR. Any data you can cite would be helpful.

yury-s:
> Can you speak more about the negative impact of including the support that led to this reversion?

The main reason "is not covert by tests yet", we claim support for a new platform if we don't run a single test for that and don't have a bot checking that it doesn't break. We also want to add arm support for mac and that means bundling 2 more drivers in addition to 3 today. At this point we'll likely split the driver bundle into several platform-specific ones.

otterley:
Is there anything I can do to assist with the testing component? I have access to ARM64 Linux resources to contribute to this effort.

niraj-khandelwal:
I do host lot of clients (some of them Fortune500) companies. aarch64 is a great architecture and an impressive one. My management is forcing to move our servers to aarch64. We use playwright-java for one of our critical functionality. This is really blocking our movement.

Requesting you to include aarch64 in the earliest release.

vcrego:
This is one request we're waiting, we are all set with out arm64 setup. Some functionality in our ppm application use playwright-java and due to no support to arm64 its failing. @yury-s - Request you to help and move this on priority so that will help us move fast here. @otterley Thanks for putting this together. You guys are the Best.!!

steveseaney:
We have over close to a thousand environments that will significantly benefit from this change.   What can we do to assist with testing?  The change prevents us from implementing a significant cost reduction.

otterley:
@yury-s 

1. AFAICT, all of the platform-specific code is in the https://github.com/microsoft/playwright repo, where they're already testing for Linux/arm64 compatibility. Can we not rely on their existing testing? Is it not safe to assume that if it works, that this Java wrapper will also work on the same architecture? In what cases would it not work? Please correct me if I'm mischaracterizing the situation.

2. Since we don't know when there will be the time or resources to split the driver bundle apart, can we just proceed for the moment and optimize later? Right now the driver bundle -- even with arm64 included -- is less than 120 MiB, which is well below the noise floor for many if not most people. I'm happy to support breaking it apart as a fast follow.

PR #903, opened 2022-04-13
fix: convert file path to absolute in setInputFiles

yury-s:


playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java
```diff
@@ -215,7 +215,7 @@ public JsonArray serialize(List<KeyboardModifier> modifiers, Type typeOfSrc, Jso
   static JsonArray toJsonArray(Path[] files) {
     JsonArray jsonFiles = new JsonArray();
     for (Path p : files) {
-      jsonFiles.add(p.toString());
+      jsonFiles.add(p.toAbsolutePath().toString());
```
rwoll, line 218: Will this remove/resolve `..`? `/a/../foo`

yury-s: It will properly concatenate them with absolute path of cwd, so should be fine.

rwoll (approved):

PR #914, opened 2022-05-03
feat: add driver for mac arm64

yury-s:
#836


yury-s:
Since x64 node already works on m1 under Rosetta and we provide ship arm-specific browsers for that platform we decided not to include mac arm64 driver for now.

jrgarcia:
Is there a way to use Playwright on M1 Macs where installing Rosetta is not allowed?

yury-s:
We don't ship node.js for mac arm today. Please open a new feature request if you need it.

PR #919, opened 2022-05-09
Added Selectors and Keyboard Manipulation Example

GabrielGavrilov:
I've noticed that in the examples, there weren't any showcases on how to use selectors and keyboard manipulation, so I've created one.

Example script is simple. Playwright simply just goes to "https://playwright.dev/", searches up "getting started", clicks on the first link, and takes a screenshot of the page.

### DIFF
--- a/examples/src/main/java/org/example/SelectorsAndKeyboardManipulation.java
+++ b/examples/src/main/java/org/example/SelectorsAndKeyboardManipulation.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


examples/src/main/java/org/example/SelectorsAndKeyboardManipulation.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class SelectorsAndKeyboardManipulation {
+  public class void main(String[] args) {
+    try(Playwright playwright = Playwright.create()) {
+      Browser browser = playwright.firefox().launch();
+      BrowserContext context = browser.newContext();
+      Page page = context.newPage();
+      page.navigate("https://playwright.dev/");
+      page.click("#__docusaurus > nav > div.navbar__inner > div.navbar__items.navbar__items--right > div.searchBox_qEbK > button");
+      page.waitForSelector("#docsearch-input");
```
yury-s, line 29: You don't need any of `waitForSelector` calls before performing actions on the page, playwright will [automatically wait](https://playwright.dev/java/docs/next/actionability) for those elements to become actionable.

yury-s (commented):
Looks good overall, but since we want to add it as an official recommendation I'd like the code to be at least free of the antipatterns like waitForSelector before .click etc.

PR #920, opened 2022-05-09
Added Selectors and Keyboard Manipulation Example

GabrielGavrilov:
I've noticed that in the examples, there weren't any showcases on how to use selectors and keyboard manipulation, so I've created one.

One of the main things that makes Playwright "Playwright" and just a web scraper is keyboard manipulation. One of the main things that Playwright can do that other web scrapers can't, is click on different selectors and manipulate the keyboard into accessing certain pages. 

I've made a simple example of how Playwright works with selectors. This example is quite simple. All it does is just go to "https://playwright.dev/java/", searches up "getting started" clicks on the First Script button, and takes a screenshot once new page loads.

### DIFF
--- a/examples/src/main/java/org/example/SelectorsAndKeyboardManipulation.java
+++ b/examples/src/main/java/org/example/SelectorsAndKeyboardManipulation.java
@@ -24,14 +24,12 @@ public class SelectorsAndKeyboardManipulation {
       Browser browser = playwright.firefox().launch();
       BrowserContext context = browser.newContext();
       Page page = context.newPage();
-      page.navigate("https://playwright.dev/");
-      page.click("#__docusaurus > nav > div.navbar__inner > div.navbar__items.navbar__items--right > div.searchBox_qEbK > button");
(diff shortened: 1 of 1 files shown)


yury-s (approved):

examples/src/main/java/org/example/SelectorsAndKeyboardManipulation.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.example;
+
+import com.microsoft.playwright.*;
+
+public class SelectorsAndKeyboardManipulation {
+  public class void main(String[] args) {
```
MichalAnczkiewicz, line 22: Shouldn't there be `static` instead of `class`?

GabrielGavrilov: I'm so embarrassed! I Can't believe I missed that. I'll change it right now.

PR #974, opened 2022-06-28
Artifact should have same version as release.

gemiusz:


yury-s (commented):
Looks good but let's wait for the release to be published.

gemiusz:
@yury-s I have looking around repo and I wondering if setting correct version should be added to:
`https://github.com/microsoft/playwright-java/blob/main/scripts/update_readme.sh`

yury-s:
We update_readme.sh on both release branches and main. I don't think we want to update the project with unreleased version when we run on main and on the other hand it does not make sense to update the project and readme on the release branches as the users normally see and use main content. What we actually want is to update README and examples' pom.xml on main with the latest released version which I think can be automatically detected by looking at the version tags.

PR #1003, opened 2022-07-22
chore(roll): roll 1.24.0 add regex, date, url serializers

mxschmitt:
https://github.com/microsoft/playwright-java/issues/987


playwright/src/main/java/com/microsoft/playwright/impl/LocatorImpl.java
```diff
@@ -67,7 +67,8 @@ public LocatorImpl(FrameImpl frame, String selector, LocatorOptions options) {
       if (options.hasText != null) {
         if (options.hasText instanceof Pattern) {
           Pattern pattern = (Pattern) options.hasText;
-          selector += " >> :scope:text-matches(" + escapeWithQuotes(pattern.pattern()) + ", \"" + toJsRegexFlags(pattern) + "\")";
+          String jsRegex = "/" + pattern.pattern() + "/" + toJsRegexFlags(pattern);
+          selector += " >> has=" + gson().toJson("text=" + jsRegex);
```
rwoll, line 71: Does this match the others? I think the gson encoding should just be around jsRegex. Might not matter, but match whichever JS does

debs-obrien (approved):

rwoll (approved):

playwright/src/main/java/com/microsoft/playwright/impl/LocatorImpl.java
mxschmitt: yes same as others.

playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java
```diff
@@ -144,6 +148,14 @@ private SerializedValue serializeValue(Object value) {
         result.n = (Integer) value;
       } else if (value instanceof String) {
         result.s = (String) value;
+      } else if (value instanceof Date) {
+        result.d = ((Date)value).toInstant().toString();
+      } else if (value instanceof URL) {
+        result.s = ((URL)value).toString();
```
yury-s, line 154: this should be `result.u = ((URL)value).toString();` ? Let's add a test for this.

playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java
```diff
@@ -207,6 +219,17 @@ private static <T> T deserialize(SerializedValue value, Map<Integer, Object> idT
       return (T) value.b;
     if (value.s != null)
       return (T) value.s;
+    if (value.u != null) {
+      try {
+        return (T)(new URL(value.u));
+      } catch (MalformedURLException e) {
+        return null;
```
yury-s, line 226: Don't swallow the exception, `throw new PlaywrightException("Unexpected value: " + value.u, e);`

playwright/src/test/java/com/microsoft/playwright/TestPageEvaluate.java
```diff
@@ -574,20 +579,52 @@ void shouldEvaluateException2() {
...
+  @Test
+  void shouldEvaluateUrl() throws MalformedURLException {
+    Object result = page.evaluate("() => ({ url: new URL('https://example.com/') })");
+    assertEquals(mapOf("url", new URL("https://example.com/")), result);
+  }
+
+  @Test
+  void shouldRoundtripUrl() throws MalformedURLException {
+    URL url = new URL("https://example.com");
+    Object result = page.evaluate("url => url", url);
```
yury-s, line 619: This also needs to check that the result is an instance of URL.

yury-s (approved):

yury-s (commented):
TestPageScreenshot.shouldCaptureBlinkingCaretIfExplicitlyAskedFor is failing in FF across platforms, we should update the expectations.

yury-s (approved):

PR #1038, opened 2022-08-25
tests: use assertThrows instead of try-catch

Meir017:
I wasn't sure what to do about this one in `TestPdf.java`
looks like it only runs on non-chromium
```java
  @Test
  @DisabledIf(value="com.microsoft.playwright.TestBase#isChromium", disabledReason="skip")
  void shouldOnlyHavePdfInChromium() {
    try {
      page.pdf();
      if (isChromium()) {
        return;
      }
      fail("did not throw");
    } catch (PlaywrightException e) {
      assertFalse(e.getMessage().contains("did not throw"));
    }
  }
```

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestPdf.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestPdf.java
@@ -52,6 +52,6 @@ public class TestPdf extends TestBase {
   @DisabledIf(value="com.microsoft.playwright.TestBase#isChromium", disabledReason="skip")
   void shouldThrowInNonChromium() {
     PlaywrightException e = assertThrows(PlaywrightException.class, () -> page.pdf());
-    assertFalse(e.getMessage().contains("Page.pdf only supported in headless Chromium"));
+    assertTrue(e.getMessage().contains("Page.pdf only supported in headless Chromium"));
   }
 }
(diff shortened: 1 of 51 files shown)


yury-s (approved):
Nice! Thanks for doing this!

yury-s:
> I wasn't sure what to do about this one in `TestPdf.java`
> looks like it only runs on non-chromium

Yes, it only runs on Chromium so we should just remove isChromium check and do assertThrows. The test can be renamed to shouldThrowInNonChromium or something like that.

Meir017:
@yury-s Done

playwright/src/test/java/com/microsoft/playwright/TestPdf.java
```diff
@@ -50,16 +50,8 @@ void shouldSupportFractionalScaleValue(@TempDir Path tempDir) throws IOException
...
-      if (isChromium()) {
-        return;
-      }
-      fail("did not throw");
-    } catch (PlaywrightException e) {
-      assertFalse(e.getMessage().contains("did not throw"));
-    }
+  void shouldThrowInNonChromium() {
+    PlaywrightException e = assertThrows(PlaywrightException.class, () -> page.pdf());
+    assertFalse(e.getMessage().contains("did not throw"));
```
yury-s, line 55: this is not necessary any more, let's replace it with a check for the actual message.

yury-s (approved):

yury-s:
I can merge it once the compilation error is fixed.

playwright/src/test/java/com/microsoft/playwright/TestPdf.java
Meir017: Done

Meir017:
pipeline run https://github.com/Meir017/playwright-java/actions/runs/2940207866

yury-s (approved):

PR #1042, opened 2022-08-31
test: use assertThrows instead of try-catch - remaining tests

Meir017:


playwright/src/test/java/com/microsoft/playwright/TestPageScreenshot.java
```diff
@@ -111,12 +111,7 @@ void shouldResumeInfiniteAnimations() {
     byte[] buffer1 = page.screenshot();
     rafraf(page);
     byte[] buffer2 = page.screenshot();
-    try {
-      assertArrayEquals(buffer1, buffer2);
-    } catch (AssertionFailedError e) {
-      return;
-    }
-    fail("Screenshots are equal");
+    assertThrows(AssertionFailedError.class, () -> assertArrayEquals(buffer1, buffer2));
```
Meir017, line 114: we are basically doing an `assertArrayNotEquals`

playwright/src/test/java/com/microsoft/playwright/TestRequestContinue.java
```diff
@@ -75,18 +75,14 @@ void shouldOverrideRequestUrl() throws ExecutionException, InterruptedException
...
-      try {
+      PlaywrightException e = assertThrows(PlaywrightException.class, () -> {
         route.resume(new Route.ResumeOptions().setUrl("file:///tmp/foo"));
-      } catch (PlaywrightException e) {
-        error[0] = e;
-        route.resume();
-      }
+      });
+      assertTrue(e.getMessage().contains("New URL must have same protocol as overridden URL"), e.getMessage());
+      route.resume();
```
Meir017, line 83: I think this is correct due to the route handler being executed only once and not multiple times - previously was expecting at least one exceptions for all interceptions, now for each intercept we except an exception

playwright/src/test/java/com/microsoft/playwright/TestTap.java
```diff
@@ -103,11 +104,7 @@ void shouldWaitForANavigationCausedByATap() throws InterruptedException {
     List<String> events = Collections.synchronizedList(new ArrayList<>());
     server.setRoute("/intercept-this.html", exchange -> {
       // make sure the tap doesnt resolve too early
-      try {
-        Thread.sleep(100);
-      } catch (InterruptedException e) {
-        events.add("interrupted");
-      }
+      assertDoesNotThrow(() -> Thread.sleep(100));
```
Meir017, line 107: based on the fact that the assert at the end of the test does not expect `"interrupted"` to be in the array we are actually expecting this call to not throw

playwright/src/test/java/com/microsoft/playwright/TestPageEvaluate.java
```diff
@@ -532,16 +532,15 @@ void shouldWorkWithCSP() {
 
   @Test
   void shouldEvaluateException() {
-    try {
+    PlaywrightException e = assertThrows(PlaywrightException.class, () -> {
```
Meir017, line 535: this test is failing, looks like it used to pass due to the test not verifying that an exception is thrown

playwright/src/test/java/com/microsoft/playwright/TestPageEvaluate.java
```diff
@@ -532,16 +532,15 @@ void shouldWorkWithCSP() {
...
       page.evaluate("() => {\n" +
         "  return (function functionOnStack() {\n" +
         "    return new Error('error message');\n" +
```
yury-s, line 538:
```suggestion
        "    throw new Error('error message');\n" +
```
This should fix the test.

playwright/src/test/java/com/microsoft/playwright/TestPageEvaluate.java
yury-s: Yeah, looks like a typo, see my suggestion below how to fix it.

playwright/src/test/java/com/microsoft/playwright/TestRequestContinue.java
yury-s: Sounds good.

yury-s (approved):
Awesome!

playwright/src/test/java/com/microsoft/playwright/TestScreencast.java
```diff
@@ -76,15 +76,13 @@ void saveAsShouldThrowWhenNoVideoFrames(@TempDir Path videosDir) {
...
-      } catch (PlaywrightException e) {
-        // WebKit pauses renderer before win.close() and actually writes something.
-        if (isWebKit()) {
-          assertTrue(Files.exists(saveAsPath));
-        } else {
-          assertTrue(e.getMessage().contains("Page did not produce any video frames"), e.getMessage());
-        }
+        assertTrue(Files.exists(saveAsPath));
+      } else {
+        PlaywrightException e = assertThrows(PlaywrightException.class, () -> popup.video().saveAs(saveAsPath));
```
yury-s, line 84: The test [failed](https://github.com/microsoft/playwright-java/runs/8121694827?check_suite_focus=true) on one of the bots:
image

Looks like there is a race between popup closure and us calling saveAs on the video. We may need to wait for the popup to finish closing?

```java
if (!popup.isClosed()) {
    popup.waitForClose(() -> {});
}
```

Meir017: done

playwright/src/test/java/com/microsoft/playwright/TestPageEvaluate.java
Meir017: only webkit is failing now, an exception is not being thrown here

yury-s: [Upstream test](https://github.com/microsoft/playwright/blob/27ffdcc9440d9d76ddca1c8059b46ee98c42ac25/tests/page/page-evaluate.spec.ts#L575-L583) does not actually throw, it returns an error and checks that it is properly formatted. Let me align java test with the upstream one.

yury-s (approved):

PR #1046, opened 2022-09-06
tests: remove duplicate cookies test

Meir017:
follow up on #1045

remove duplicate test `TestBrowserContextAddCookies.shouldRoundtripCookie` of `TestBrowserContextCookies.shouldGetANonSessionCookie`

the original test is `browsercontext-cookies.spec.ts` - `should get a non-session cookie`

Maybe we should annotate the test-classes and test methods with the `@DisplayName` similar to how the `playwright-dotnet` repo does?

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestBrowserContextAddCookies.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestBrowserContextAddCookies.java
@@ -39,6 +39,24 @@ public class TestBrowserContextAddCookies extends TestBase {
   }
 
   @Test
+  void shouldRoundtripCookie() {
+    page.navigate(server.EMPTY_PAGE);
+    // @see https://en.wikipedia.org/wiki/Year_2038_problem
+    Object documentCookie = page.evaluate("() => {\n" +
+      "  const date = new Date('1/1/2038');\n" +
+      "  document.cookie = `username=John Doe;expires=${date.toUTCString()}`;\n" +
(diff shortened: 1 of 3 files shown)


yury-s:
> Maybe we should annotate the test-classes and test methods with the `@DisplayName` similar to how the `playwright-dotnet` repo does?

Yes, we can start doing this at least for the new tests. It doesn't have to be @DisplayName,  having just a permalink in the comment to the original test would work.

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextAddCookies.java
```diff
@@ -40,34 +38,6 @@ void shouldWork() {
     assertEquals("password=123456", page.evaluate("document.cookie"));
   }
 
-  @Test
-  void shouldRoundtripCookie() {
```
yury-s, line 44: This test is actually present upstream and seems that https://github.com/microsoft/playwright-java/pull/1045/ actually changed the tests logic a bit which makes it now indistinguishable from shouldGetANonSessionCookie. Let's align this test with upstream instead of deleting it:

```java
    List<Cookie> cookies = context.cookies();
    assertEquals(1, cookies.size());
    context.clearCookies();
    assertEquals(emptyList(), context.cookies());
    context.addCookies(cookies);
    assertJsonEquals(new Gson().toJson(cookies), context.cookies());```

PR #1064, opened 2022-09-16
Added example in README.md for Gradle project

pranesh517:
Added example in README.md to add Playwright dependency in build.gradle


yury-s (approved):
You may also want to add a Gradle example to https://playwright.dev/java/docs/intro, the source is at  https://github.com/microsoft/playwright/blob/main/docs/src/intro-java.md

PR #1142, opened 2022-12-16
feat: roll driver, implement new APIs

yury-s:
https://github.com/microsoft/playwright-java/issues/1127


dgozman (approved):

playwright/src/main/java/com/microsoft/playwright/APIRequestContext.java
```diff
@@ -384,7 +396,7 @@ default APIResponse put(String url) {
   }
   /**
    * Sends HTTP(S) <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT">PUT</a> request and returns its
-   * response. The method will populate request cookies from the context and update context cookies from the response. The
+   * response.↵The method will populate request cookies from the context and update↵context cookies from the response. The
```
mxschmitt, line 399:
```suggestion
   * response. The method will populate request cookies from the context and update↵context cookies from the response. The
```

You need to replace them with spaces.

playwright/src/main/java/com/microsoft/playwright/Frame.java
```diff
@@ -1410,80 +1412,64 @@ public IsEnabledOptions setTimeout(double timeout) {
   }
   class IsHiddenOptions {
     /**
-     * When true, the call requires selector to resolve to a single element. If given selector resolves to more than one
+     * When true, the call requires selector to resolve to a single element. If given selector resolves to more↵than one
      * element, the call throws an exception.
      */
     public Boolean strict;
-    /**
-     * **DEPRECATED** This option is ignored. {@link Frame#isHidden Frame.isHidden()} does not wait for the element to become
```
mxschmitt, line 1418: you don't render deprecated method reasons anymore, its now deprecated: string instead of deprecated: bool and contains the reason.

playwright/src/main/java/com/microsoft/playwright/APIRequestContext.java
yury-s: This is something new!

PR #1165, opened 2023-01-10
feat: Upgrade to Java 11 and module system

palexdev:
Make Playwright finally easy to use in modular projects

Oh, this will also require the update of workflows to use at least JDK11

### DIFF
--- a/driver-bundle/pom.xml
+++ b/driver-bundle/pom.xml
@@ -4,9 +4,9 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>com.microsoft.playwright</groupId>
+    <groupId>io.github.palexdev.playwright</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>1.29.0-SNAPSHOT</version>
+    <version>1.30.0</version>
   </parent>
 
   <artifactId>driver-bundle</artifactId>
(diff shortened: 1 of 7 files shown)


palexdev:

driver-bundle/pom.xml
```diff
@@ -37,9 +37,9 @@
   </build>
   <dependencies>
     <dependency>
-      <groupId>com.microsoft.playwright</groupId>
+      <groupId>io.github.palexdev.playwright</groupId>
```
yury-s, line 40: Revert this and similar changes in other pom.xml files

driver-bundle/pom.xml
```diff
@@ -37,9 +37,9 @@
...
       <artifactId>driver</artifactId>
-      <version>${project.version}</version>
+      <version>1.30.0</version>
```
yury-s, line 42: ditto

driver-bundle/pom.xml
palexdev: Oh I'm very sorry for this, I didn't know the commit would be automatically added to the PR
Since I needed the library in one of my modular projects yesterday I decided to publish artifacts on my maven repo
I'll update the PR in the morning

driver-bundle/pom.xml
palexdev: Same issue of above sorry

pom.xml
```diff
@@ -42,7 +42,7 @@
   </modules>
 
   <properties>
-    <compiler.version>1.8</compiler.version>
+    <compiler.version>11</compiler.version>
```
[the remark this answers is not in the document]
palexdev: No it won't work with Java 8 anymore
The mode system has been introduced by Java 9, the closest LTS release is JDK11

PR #1166, opened 2023-01-12
feat: Upgrade to Java 11 and module system

palexdev:
Make Playwright finally easy to use in modular projects

This will also require the update of workflows to use at least JDK11

@yuri-s This PR is the right one, fixes the one I closed #1165

### DIFF
--- a/driver-bundle/src/main/java/module-info.java
+++ b/driver-bundle/src/main/java/module-info.java
@@ -1,0 +1,7 @@
+module playwright.driver.bundle {
+  requires playwright.driver;
+
+  exports com.microsoft.playwright.impl.driver.jar;
+
+  opens com.microsoft.playwright.impl.driver.jar;
+}
--- a/driver/src/main/java/module-info.java
+++ b/driver/src/main/java/module-info.java
@@ -1,0 +1,3 @@
+module playwright.driver {
+  exports com.microsoft.playwright.impl.driver;
+}
--- a/playwright/src/main/java/module-info.java
+++ b/playwright/src/main/java/module-info.java
@@ -1,0 +1,14 @@
+module playwright {
+  requires playwright.driver;
+  requires jdk.httpserver;
+  requires java.desktop;
+  requires com.google.gson;
+  requires org.opentest4j;
+
+  exports com.microsoft.playwright.assertions;
+  exports com.microsoft.playwright.impl;
+  exports com.microsoft.playwright.options;
+
+  opens com.microsoft.playwright;
+  opens com.microsoft.playwright.impl;
+}
--- a/playwright/src/main/java/module-info.java
+++ b/playwright/src/main/java/module-info.java
@@ -8,6 +8,7 @@ module playwright {
   exports com.microsoft.playwright.assertions;
   exports com.microsoft.playwright.impl;
   exports com.microsoft.playwright.options;
+  exports com.microsoft.playwright;
 
   opens com.microsoft.playwright;
   opens com.microsoft.playwright.impl;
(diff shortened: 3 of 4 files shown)


yury-s:
> This will also require the update of workflows to use at least JDK11

According to [your comment](https://github.com/microsoft/playwright-java/pull/1165#discussion_r1066458923) from the original PR it will break Java 8 clients. According to the [JetBrains' survay](https://www.jetbrains.com/lp/devecosystem-2021/java/#Java_which-versions-of-java-do-you-regularly-use) most of the people are still on Java 8, we cannot break them at the moment. We need to find a way to add module support without breaking existing clients.

palexdev:
According to the lastest [JetBrains Statistics](https://www.jetbrains.com/lp/devecosystem-2022/java/) Java 8 is still predominant but its usage is rapidly decreasing in favor of JDK 11 and 17

I'd also like to find a way to add modules support because currently it's almost impossible to use Playwright in such projects. The first solution that comes to my mind would be to have a separate artifact for JDK11+, but from a "maintaining" perspective I don't know how much feasible it is

GuillaumeTaffin:
Hi, I would add some points that I find interesting on this topic

1. A vast majority of Java devs use Spring/Spring Boot
According to the stats you shared, we see that it's obviously the case.
Well, Spring Boot 3 was release at the end of 2022 and it enforces the use of Java 17. Hence I personally think that it will encourage even more organizations to move forward to latest versions of the JDK. 

2. Playwright is a relatively new player
According to the release dates I just looked, playwright-java began in 2020. 
This is a personal thought, but I think that organizations that are still stuck with java 8 today are more likely to be companies that already had their "UI testing suite" long before playwright came into the game. I feel it's less likely that that kind of organization decided to move to playwright.
In other words, organizations that choose playwright might probably be the "good students" of that world, that try new technologies and stay up-to-date. This means that they might have already migrated - or even started - to new versions of the JDK. 
Once again, this is my personal understanding of the java world - where I daily worked in a big banking company that tries to stay up-to-date - and this would need to be measured against the set of "real users of playwright-java".
I think it's important to keep in mind that the Jetbrains's stats measure the "java community" not the "playwright-java users community". The figures could be quite different for that particular subset of people IMHO

anthonyvdotbe:
It's possible to have both, module descriptors and compatibility with Java SE 8, by using multi-release JARs: you compile all sources with Java SE 8 and compile the module descriptors with Java SE 9+. See e.g. slf4j's [configuration](https://github.com/qos-ch/slf4j/blob/master/pom.xml#L184) and the [module descriptor for slf4j-api](https://github.com/qos-ch/slf4j/blob/master/slf4j-api/src/main/java9/module-info.java) (note that it's under `src/main/java9` instead of `src/main/java`).

blutorange:
Until this gets resolved properly, can we add a  `Automatic-Module-Name` to the manifest, see http://branchandbound.net/blog/java/2017/12/automatic-module-name/ 

Without an explicit name, the compiler will warn when attempting to use this library. That manifest entry is safe and you can also reserve the module name for this project.

E.g.

```xml
<configuration>
  <archive>
    <manifestEntries>
      <Automatic-Module-Name>playwright.driver.bundle</Automatic-Module-Name>
    </manifestEntries>
  </archive>
</configuration>

helpermethod:
Hi,

just to chime in: You can support Java 8 while still support Java 9 modules **without** Multi-Release JARs by using the Moditect Maven Plugin

Usage is quite simple. Here's an example:

https://github.com/helpermethod/zip-forge/blob/6ed82db3688b9d493abacc72cd26dff7e3f91b5a/pom.xml#L213

mxschmitt:
I'll close it for now since it looks stale.

PR #1187, opened 2023-01-24
Fix code snippet javadoc of Playwright.create method

chamu228:


playwright/src/main/java/com/microsoft/playwright/Playwright.java
```diff
@@ -97,7 +97,7 @@ public CreateOptions setEnv(Map<String, String> env) {
    * Launches new Playwright driver process and connects to it. {@link Playwright#close Playwright.close()} should be called
    * when the instance is no longer needed.
    * <pre>{@code
-   * Playwright playwright = Playwright.create()) {
+   * Playwright playwright = Playwright.create();
```
yury-s, line 100: This code is autogenerated. The fix should go upstream, this particular snippet is produced from https://github.com/microsoft/playwright/blob/1f576ff0bc176e3f7952971de86a370c7acd2dec/docs/src/api/java.md?plain=1#L58, please send your fix there.

chamu228: Thanks for the information. Raised pull request https://github.com/microsoft/playwright/pull/20356

mxschmitt: Merged, it gets automatically rolled into Java 1.31 then. Closing by that this one here.

PR #1188, opened 2023-01-24
chore(examples): add timeouterror exception

sohbat17:
- Add TimeOutError reference example while locating text 
- Print the message for explicitly for TimeOutError
- Print the message for general exception


sohbat17:

yury-s (commented):
I'm not sure how useful this example is as it doesn't  demonstrate anything that is not already mentioned in the API docs directly and timeout is a pretty simple concept not specific to playwright. What is the rational for adding it?

yury-s:
Closing as part of the triage process since it seems stale.

PR #1189, opened 2023-01-24
fix(timeout): add additional validation checks for the timeout

sohbat17:
- Check if the timeout is positive number before setting the timeout configuration.
- Check if the timeout is positive number before setting the navigationTimeout configuration

### DIFF
--- a/driver-bundle/pom.xml
+++ b/driver-bundle/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.microsoft.playwright</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>1.29.0-SNAPSHOT</version>
+    <version>1.30.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>driver-bundle</artifactId>
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.microsoft.playwright</groupId>
     <artifactId>parent-pom</artifactId>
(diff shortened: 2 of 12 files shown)


mxschmitt:
How can it ever be negative?

yury-s (commented):
What user observable problem does this fix?

srivatsa-cfp:
To avoid user settings negative values for timeout in PageImpl & BrowserContextImpl class.

sohbat17:
To avoid user passing negative values for timeout in PageImpl and BrowserContextImpl class.

sohbat17:
> How can it ever be negative?

    - In PageImpl & BrowserContextImpl class, there are 2 methods where timeout values are set. To avoid setting negative timeout values from these methods additional checks are added.

yury-s:
> To avoid user passing negative values for timeout in PageImpl and BrowserContextImpl class.

I understand that, what I'm asking is whether setting negative value causes any issues in the behavior of the program?

sohbat17:
> > To avoid user passing negative values for timeout in PageImpl and BrowserContextImpl class.
> 
> I understand that, what I'm asking is whether setting negative value causes any issues in the behavior of the program?

Setting negative will stop execution of statements and test execution flow. for example we usually keep a certain duration for the async user operations like waiting for response, dynamic page locator based on previous selection etc.

yury-s:
Same will happen if one sets timeout to 0 or actually to anything close enough to 0.  Doesn't seem to make any difference from the user's stand point once the timeout is less than a few milliseconds - the behavior will be all the same.

PR #1200, opened 2023-02-13
fix: Improve maven plugin configuration

### LINKED ISSUE #1199: [Feature] - Improve maven plugin configuration
I recommend to use more recent version of maven-compiler-plugin and use the default properties which makes configuration easiert.

khmarbaise:
Fixes #1199

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,8 @@
   </modules>
 
   <properties>
-    <compiler.version>1.8</compiler.version>
+    <maven.compiler.source>8</maven.compiler.source>
+    <maven.compiler.target>8</maven.compiler.target>
     <gson.version>2.8.9</gson.version>
     <junit.version>5.7.0</junit.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


yury-s (approved):
Just curious, are there any other advantages except for use of predefined property names?

pom.xml
```diff
@@ -42,7 +42,8 @@
   </modules>
 
   <properties>
-    <compiler.version>1.8</compiler.version>
+    <maven.compiler.source>8</maven.compiler.source>
```
yury-s, line 45: Does it actually support 8 interchangeably with 1.8? I can find it immediately in the docs and all examples seem to use 1.8

yury-s: Ok, it's just [-source](https://github.com/apache/maven-compiler-plugin/blob/b225b1c29b1ee2fc2b178d2f2f1b920afe1d9a75/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java#L180) compiler option and from the javac help:

```
  --source <release>, -source <release>
        Provide source compatibility with the specified Java SE release. Supported releases: 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18
```
so it should be good.

khmarbaise: Those properties will be automatically picked up by other plugins  (for example the [maven-javadoc-plugin](https://maven.apache.org/plugins/maven-javadoc-plugin/jar-mojo.html#source)) while the `compiler.version` is not...and yes you can use `8` instead of `1.8` (related to the docs that looks like a lack of detailed docs..)...also so the ran builds show that...

Issue #1201, opened 2023-02-13
[Feature] - Improve plugin configuration and use inheritance


khmarbaise:
Currently a lot of configuration is repeated in sub module like `driver-bundle`, `playwright` etc. repeating all plugins just by using 
```xml
 <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-source-plugin</artifactId>
      </plugin>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-javadoc-plugin</artifactId>
      </plugin>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
      </plugin>
    </plugins>
  </build>
```
which can simply being handled by usign inheritance. Not defined all plugins version to make sure each time the build will be executed to use the same versions of plugins.

PR #1202, opened 2023-02-13
devops: Improve plugin configuration inheritance usage (#1201)

khmarbaise:
devops: Improve plugin configuration inheritance usage.
     - removing duplication of plugin definition in modules
     - using single location to define all (parent)
     - defined all used plugins via pluginManagement with
       most recent versions.
     - removed life cycle binding from pluginManagement
     - moved life cycle binding to default location
     - using jar-no-fork of maven-source-plugin.

### DIFF
--- a/driver-bundle/pom.xml
+++ b/driver-bundle/pom.xml
@@ -16,25 +16,6 @@
     It is intended to be used on the systems where Playwright driver is not preinstalled.
   </description>
 
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-source-plugin</artifactId>
-        <configuration>
-          <excludeResources>true</excludeResources>
-        </configuration>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-javadoc-plugin</artifactId>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-      </plugin>
-    </plugins>
-  </build>
   <dependencies>
     <dependency>
       <groupId>com.microsoft.playwright</groupId>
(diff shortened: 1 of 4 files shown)


driver-bundle/pom.xml
```diff
@@ -16,25 +16,6 @@
     It is intended to be used on the systems where Playwright driver is not preinstalled.
   </description>
 
-  <build>
-    <plugins>
```
yury-s, line 20: With `<plugins>` section removed in the child projects, how do we ensure that corresponding plugins are executed for each of them? It seems to work but I'd like to understand why as according to the [pluginManagement](https://maven.apache.org/pom.html#plugin-management) page: "However, this only configures plugins that are actually referenced within the plugins element in the children or in the current POM. " which I interpret (perhaps wrongly) as the plugins still have to be added to `<plugins>` section.

khmarbaise: The plugins are bound by the life cycle for jar, war etc. (https://maven.apache.org/ref/current/maven-core/default-bindings.html) that means it is only necessary to define their versions via. pluginManagement but not explicitly mention them or bind them... for example the maven-surefire-plugin is bound by default (reference above) which means only versions/configuration change in the pluginManagement. Exception of them for example the maven-source-plugin because it's not bound to the life cycle in anyway... that's the reason to bind it manually via `<executions>..</executions>` in the parent which is inherited into the childs... That's the reason to define version/configuration in pluginManagement block but added the binding explictly in `<build><plugins><plugin>...</plugin></plugins></build>` area...also the maven-javadoc-plugin... Also if you would use the maven-failsave-plugin this is not bound by default so you have to add a binding...(https://maven.apache.org/surefire/maven-failsafe-plugin/usage.html) ...

khmarbaise: Ah... the maven-jar-plugin is bound by default but the goal `test-jar` is not bound...that's the reason why it's needed to do that explicitly... if you like to create jar of the test parts... for reusing..(not always the best way)..

yury-s: Thanks for explaining! I'm not a big fun of implicit default plugins but the configuration is indeed less verbose, so let's give it a try!

yury-s (approved):

Issue #1213, opened 2023-02-22
Not able to use system node JS path  - C:\Program is not recognized as an internal or external command


mEklavya:
- Playwright Version:1.30.0
- Operating System: Windows
- Browser: Tried for Chrome

Java

try (Playwright playwright = Playwright.create()) {
			Browser browser = playwright.chromium().launch(
		    		  new BrowserType.LaunchOptions().setHeadless(false)
		    		  .setExecutablePath(Paths.get("C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe"))
		    		  );
		     
		      Page page1 = browser.newPage();
		      page1.navigate("http://playwright.dev");
		      System.out.println(page1.title());
}
Setting up the NODE JS PATH with the below env property
PLAYWRIGHT_NODEJS_PATH=C:\Program Files\nodejs
Also used below property
PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1

Code works fine if use existing nodeJS, if we set PLAYWRIGHT_NODEJS_PATH then we are getting below error
'""C:\Program' is not recognized as an internal or external command,
operable program or batch file.
Exception in thread "main" com.microsoft.playwright.PlaywrightException: Failed to read message from driver, pipe closed.
	at com.microsoft.playwright.impl.PipeTransport.poll(PipeTransport.java:75)
	at com.microsoft.playwright.impl.Connection.processOneMessage(Connection.java:175)
	at com.microsoft.playwright.impl.ChannelOwner.runUntil(ChannelOwner.java:109)
	at com.microsoft.playwright.impl.Connection.sendMessage(Connection.java:119)
	at com.microsoft.playwright.impl.ChannelOwner.sendMessage(ChannelOwner.java:102)
	at com.microsoft.playwright.impl.Connection$Root.initialize(Connection.java:79)
	at com.microsoft.playwright.impl.Connection.initializePlaywright(Connection.java:152)
	at com.microsoft.playwright.impl.PlaywrightImpl.createImpl(PlaywrightImpl.java:52)
	at com.microsoft.playwright.impl.PlaywrightImpl.create(PlaywrightImpl.java:35)
	at com.microsoft.playwright.Playwright.create(Playwright.java:110)
	at com.microsoft.playwright.Playwright.create(Playwright.java:114)

Add any other details about the problem here.
![image](https://user-images.githubusercontent.com/42370888/220539223-625f0553-b655-45ad-bd47-b1436f4c2af5.png)

[...]

PR #1223, opened 2023-03-01
chore: roll 1.32.0-alpha driver

yury-s:
#1208
Fixes https://github.com/microsoft/playwright-java/issues/1213


playwright/src/main/java/com/microsoft/playwright/impl/Connection.java
```diff
@@ -67,6 +68,7 @@ public class Connection {
   }
   LocalUtils localUtils;
   final Map<String, String> env;
+  private Set<List<CallMetadata>> stackCollectors = Collections.newSetFromMap(new IdentityHashMap<>());
```
dgozman, line 71: This is some involved construction 😄

dgozman (approved):

playwright/src/main/java/com/microsoft/playwright/impl/Connection.java
yury-s: Yeah, it's java...

PR #1248, opened 2023-03-27
chore: throw on context.close() if it was closed externally

yury-s:
This is a backport of https://github.com/microsoft/playwright/commit/09ff7eaaf2ab40fcaf0abb85f0308a91e1460a97

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/BrowserContextImpl.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/BrowserContextImpl.java
@@ -47,7 +47,8 @@ class BrowserContextImpl extends ChannelOwner implements BrowserContext {
   private final APIRequestContextImpl request;
   final List<PageImpl> pages = new ArrayList<>();
   final Router routes = new Router();
-  private boolean isClosedOrClosing;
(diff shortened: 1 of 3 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/BrowserContextImpl.java
```diff
@@ -47,7 +47,8 @@ class BrowserContextImpl extends ChannelOwner implements BrowserContext {
   private final APIRequestContextImpl request;
   final List<PageImpl> pages = new ArrayList<>();
   final Router routes = new Router();
-  private boolean isClosedOrClosing;
+  private boolean closeWasCalled;
+  private final WaitableEvent<EventType, ?> closePromise;
```
mxschmitt, line 51: `?` is the Python's type Any?

mxschmitt (commented):
Looks different to the upstream impl. I'd if closeWasCalled and then early return to not have to wait for the closePromise.

yury-s:
> Looks different to the upstream impl. I'd if closeWasCalled and then early return to not have to wait for the closePromise.

There is a bug in the upstream implementation, it should wait for close promise.

PR #1285, opened 2023-05-23
Updated dead links to use Playwright homepage in documentation

zibdie:
Currently, the documentation uses `http://whatsmyuseragent.org/` which is now a dead link filled with ads

image

It should instead be consistant and use the Playwright homepage

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ You can find Maven project with the examples [here](./examples).
 
 #### Page screenshot
 
-This code snippet navigates to whatsmyuseragent.org in Chromium, Firefox and WebKit, and saves 3 screenshots.
+This code snippet navigates to Playwright homepage in Chromium, Firefox and WebKit, and saves 3 screenshots.
 
 ```java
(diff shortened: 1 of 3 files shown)


zibdie:
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> 
>     
>   
> 
> >

zibdie:

yury-s (approved):
That's very unfortunate. We used that page as it varies between the browsers (different user-agent) and playwright.dev is the same everywhere. But having a broken link is not good. We also have a few more in the upstream repo: https://github.com/search?q=repo%3Amicrosoft%2Fplaywright%20whatsmyuseragent.org%2F&type=code

zibdie:
@yury-s ill make PRs to correct these later today to use the playwright homepage as well

zibdie:
@yury-s As promised, I raised a PR to fix those as well: https://github.com/microsoft/playwright/pull/23250

Issue #1302, opened 2023-06-12
[Feature] Add ability to specify which browsers to install


msiemczyk:
Playwright Typescript has ability to specify which browsers to install, but Java version does not. Of course I can work around it with calling install separately and skipping download with Java. However, that would mean adding another Gradle task to build system and I rather avoid that since in my opinion framework really should have this built-in as an option for easier use.

Rather than another environment variable I think it would be nicer to have explicit option in CreateOptions; it seems like a simpler and easier thing. Internally, it is still and environment variable so the method signatures don't have to be changed.

For example:

![image](https://github.com/microsoft/playwright-java/assets/5693250/2a1b501c-0a5e-449e-b56a-7b82391d1ca1)

yury-s:
> Of course I can work around it with calling install separately and skipping download with Java. 

This is how we handle browser installation in all other languages and we'd rather switch java to the same mode. Actually we also consider replacing automatic driver extraction with a manual step (see https://github.com/microsoft/playwright-java/issues/1268 for more context) in which case it would make sense to download browsers in the same step.

msiemczyk:
Fair enough, if you think that is not desired you can reject the PR.

PR #1303, opened 2023-06-12
feat: Add ability to specify which browsers to install (#1302)

msiemczyk:
Implements feature request.


msiemczyk:
@yury-s @mxschmitt  @dgozman do I have to add a reviewer myself? how do I know who to pick? or someone just joins the PR?

mxschmitt:
After an offline discussion with Yury we decided to prefer skipping installation completely and calling maven again see https://github.com/microsoft/playwright-java/issues/1302#issuecomment-1589889528 instead of introducing more complexity. We change this part in the future most likely, so we prefer keeping it as it is for now. Thanks for your proposal tho!

PR #1327, opened 2023-07-04
fix(utils): fix globToRegex for glob list

xiaofeng-zheng:
`globToRegex` doesn't properly handle glob list with recursive globstar. For example, `{**/foo,**/bar}` should match `https://example.com/foo` and `https://example.com/bar`

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
@@ -94,7 +94,7 @@ class Utils {
   static String globToRegex(String glob) {
     StringBuilder tokens = new StringBuilder();
     tokens.append('^');
-    boolean inGroup = false;
+    int groupPos = -1;
     for (int i = 0; i < glob.length(); ++i) {
       char c = glob.charAt(i);
       if (escapeGlobChars.contains(c)) {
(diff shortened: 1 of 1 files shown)


xiaofeng-zheng:
@yury-s Could you take a look?

yury-s:
@xiaofeng-zheng can you fix the problem in the upstream repository's [glob.ts](https://github.com/microsoft/playwright/blob/36588ba98e60c86d5a8a8177ffa26eeb8acd5e2d/packages/playwright-core/src/utils/glob.ts#L20) and after that we'll ensure that the implementation is consistent across all language ports?

Implementation-wise there seems to be a bigger issue as we'd need to support nested lists like `{R*.{m*,*d},*.sh}`.

In the hindsight, supporting `{}` in the glob syntax which seems to be bash extension was a bad idea and we could confine ourselves to the minimal glob syntax. But that ship has sailed.

xiaofeng-zheng:
> 

Is it possible to import some 3P glob library?

xiaofeng-zheng:
@yury-s It's a very difficult and challenging job to support all glob features. And it seems there's no individual test case for this function in this repo. So why not use an 3P, open-source glob matcher library, like [minimatch](https://github.com/isaacs/minimatch)

yury-s:
We try to keep the implementation consistent across all language ports. The glob implementation we've been using served us quite well and didn't require much maintenance. We don't rule out usage of third-party implementation but we usually evaluate the trade-off between maintaining our own solution and using available third-party libraries, in this case it was in favor of an ad-hoc one.

xiaofeng-zheng:
Maybe I should close this PR and open an issue

PR #1362, opened 2023-08-23
chore: roll driver to 1.38.0-alpha-aug-23

yury-s:
Reference https://github.com/microsoft/playwright-java/issues/1353


playwright/src/main/java/com/microsoft/playwright/impl/BrowserContextImpl.java
```diff
@@ -710,6 +722,17 @@ protected void handleEvent(String event, JsonObject params) {
         PageImpl page = connection.getExistingObject(params.getAsJsonObject("page").get("guid").getAsString());
         page.listeners.notify(PageImpl.EventType.RESPONSE, response);
       }
+    } else if ("pageError".equals(event)) {
```
dgozman, line 725: Shouldn't similar code be removed from PageImpl now?

dgozman (approved):

PR #1366, opened 2023-08-27
feat: base JUnit classes

jfgreffier:
Adds base classes for JUnit tests, in a similar way to .Net

Suggested here https://github.com/microsoft/playwright/pull/24612#discussion_r1294948105 by Yury S

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestPageBaseClass.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestPageBaseClass.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.playwright;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+import com.microsoft.playwright.Browser.NewContextOptions;
+import com.microsoft.playwright.junit.PageTest;
+import static com.microsoft.playwright.Utils.getBrowserNameFromEnv;
+
+import static com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat;
+import static com.microsoft.playwright.options.ColorScheme.DARK;
+
+class TestPageBaseClass extends PageTest {
+
+  @Override
+  public NewContextOptions contextOptions() {
+    return new Browser.NewContextOptions().setColorScheme(DARK);
+  }
+
+  @Test
+  void shouldWork() {
+    page.navigate("data:text/html,<div>A</div>");
+    assertThat(page).hasURL("data:text/html,<div>A</div>");
+    Locator div = page.locator("div");
+    assertThat(div).containsText("A");
+  }
+
+  @Test
+  void browserNameShouldWork() {
+    assertEquals(getBrowserNameFromEnv(), browserName);
+  }
+
+  @Test
+  void shouldOverrideContextOptions() {
+    assertEquals(false, page.evaluate("matchMedia('(prefers-color-scheme: light)').matches"));
+    assertEquals(true, page.evaluate("matchMedia('(prefers-color-scheme: dark)').matches"));
+  }
+
+}
--- a/pom.xml
+++ b/pom.xml
@@ -78,6 +78,7 @@
         <groupId>org.junit.jupiter</groupId>
         <artifactId>junit-jupiter-api</artifactId>
         <version>${junit.version}</version>
+        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.junit.jupiter</groupId>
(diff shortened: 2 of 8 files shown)


playwright/src/test/java/com/microsoft/playwright/TestPageBaseClass.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.jupiter.api.Test;
+
+import com.microsoft.playwright.Browser.NewContextOptions;
+import com.microsoft.playwright.junit.PageTest;
+import static com.microsoft.playwright.Utils.getBrowserNameFromEnv;
+
+import static com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat;
+import static com.microsoft.playwright.options.ColorScheme.DARK;
+
+class TestPageBaseClass extends PageTest {
```
yury-s, line 30: We should be able to switch our own test harness to the new API and get a better coverage I think.

uchagani:
Just a suggestion, maybe we can make use of https://github.com/uchagani/junit-playwright instead?  This would allow us to offer Playwright fixtures without using inheritance and just by annotating a test class like this:

```java
@UseBrowserConfig(MyConfig.class)
public class SomeTestClass {

  void someTest(Page page) {
    Locator locator = page.getByText("foo");
    //assertion
  }  
}
```

This also provides the benefit of running ALL tests in parallel not just classes in parallel although the user can still configure per class if they wish.

If this is approach is acceptable to you @yury-s I would love to work with you @jfgreffier to build this into playwright-java itself.  In fact, if we did build it into playwright-java we could make it even better than some of the workarounds I had to make in [junit-playwright](https://github.com/uchagani/junit-playwright)

yury-s:
Let me discuss this with the team in a meeting tomorrow. We'll most likely need to get the base classes to support multi-threaded execution, but I'm not sure to what extent we want to commit to providing such framework in Java. What is in this PR can be used as a simple base implementation that could be easily replaced with more complex implementation depending on the project, whereas what you have is a more comprehensive framework that would require more thought and commitment from us.

uchagani:
sounds good.  JUnit 5 has a very robust extension system and with it's console launcher, we could make a playwright-java test runner similar to how we have playwright test, but we don't need to, we can just provide fixtures.  

My biggest concern with an inheritance based approach is that since java doesn't allow multiple inheritance, it forces the user to have 1 base class that can cause problems as a user's test suite grows.

pom.xml
```diff
@@ -73,6 +73,11 @@
         <artifactId>opentest4j</artifactId>
         <version>${opentest4j.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-api</artifactId>
+        <version>${junit.version}</version>
```
uchagani, line 79: @jfgreffier @yury-s will this force the user to use the junit version that playwright expects?  Maybe we needs some documentation about excluding it if the user runs into conflicts with the version they are using?

We make this `provided` scope but then it may not build for a user that is not using junit and using testng?

yury-s: > Maybe we needs some documentation about excluding it if the user runs into conflicts with the version they are using?

Good point! We don't Playwright to dictate which exact JUnit version must be used in the project, I agree taht `<scope>provided</scope` would be appropriate here. We'll likely have a similar package for TestNG eventually and I assume the clients will either use one or the other and will have only one of the frameworks installed (or none, in case of scraping). We should work in those scenarios.

> We could make this provided scope but then it may not build for a user that is not using junit and using testng?

Correct me if I'm wrong, but unless the user explicitly imports one of the classes that depend on junit stuff, their project should work just fine without junit being available. Ot the same time, if their code does use junit, they are supposed to have junit library added to their deps. But I agree that it will be more tricky to document than what we have today.

As an alternative, we can put JUnit (and TestNG for that matter) related classes into their own Maven module, so that clients that use junit could import something like playwright-junit and those with TestNG playwright-testng. 

Interestingly enough we've managed to avoid all this dependency issue so far with playwright assertions.

uchagani: you're correct, just did a quick test and provided scope will solve this.

yury-s:
> Just a suggestion, maybe we can make use of https://github.com/uchagani/junit-playwright instead (or use that approach and build it into playwright-java)?

@uchagani, thank you for the suggestion. We discussed this in our team meeting today and are open to incorporating your solution, although it may need some modifications as we review it. Would you be willing to contribute it to this repository? Additionally, do you feel you would have the resources to maintain it long-term?

uchagani:
@yury-s yes absolutely. I'm willing to do the bulk of the work with your guidance on changes you'd like and will help maintain it.

playwright/src/test/java/com/microsoft/playwright/TestPageBaseClass.java
jfgreffier: We should probably do it seperately, to avoid a gigantic Pull Request ?

yury-s:
> I believe we should close this to incorporate https://github.com/uchagani/junit-playwright 

Yes, if we decide to use annotations-based approach we'll need to use it everywhere including docs and codegen, which makes change in this PR obsolete.

uchagani:
@yury-s is the next step to wait for instructions from you/your team on what changes should be made to junit-playwright?

yury-s:
@uchagani, @jfgreffier I filed https://github.com/microsoft/playwright-java/issues/1369 with high-level requirements for the integration. You can start by modifying junit-playwright so that it satisfies those constraints. I believe sending PRs for the parts which are ready and reviewing/iterating on them would be the easiest path forward. We may need to remove them before the release if the feature is not ready by the release, but we can either do it in the release branch only or revert it on main and reapply later after release branch cut. We are planning to cut next release (1.38) sometime next week.

jfgreffier:
Closing as per discussion. I'm looking forward to better JUnit integration !

PR #1371, opened 2023-09-04
feat(junit-playwright)

### LINKED ISSUE #1369: [Feature] JUnit5 integration
To provide decent support of JUnit in codegen (as requested in https://github.com/microsoft/playwright-java/issues/1039), we need to first implement basic Playwright fixtures similar to other languages. This came up in the discussion of https://github.com/microsoft/playwright-java/pull/1366 and whether/how we can integrate https://github.com/uchagani/junit-playwright into Playwright. Below are some of the requirements we'll have for such integration and some comments regarding how it compares to the existing junit-playwright.

The execution model should be similar to [the default](https://playwright.dev/docs/test-parallel#parallelize-tests-in-a-single-file) execution model of Node.js Playwright. I.e. it should be possible to
 - configure browser options per test file, so that an instance of the browser is reused by all tests in the same file
 - do not parallel tests in the same file, only run in parallel tests in different classes (files), i.e. in terms of junit config:
                junit.jupiter.execution.parallel.enabled = true
                junit.jupiter.execution.parallel.mode.default = same_thread
                junit.jupiter.execution.parallel.mode.classes.default = concurrent

Builtin fixtures should create Playwright, Browser, BrowserContext and Page. Their lifetimes should match those in Node.js version, i.e.
  - new context and page are created before each test
  - for each test, same instance of page and context is used within the test, @beforeEach and @afterEach hooks 
  - page and context are closed right after last @afterEach hook finished for the test
 
Browser & playwright lifetimes should be bound to the thread & browser configuration. Playwright instance most likely should be thread local and can be reused by all tests running on that thread. We should try to reuse the same browser instance for running more tests that require the browser launch options. Currently, playwright internal test suite creates new browser per test class which seems to be good enough.

To keep things simple in the beginning we'd like to replace PlaywrightBrowserConfig with a configurable BrowserFactory, so that we don't have to think about config format and let the user create custom browser themselves. Eventually, we'll likely want to load the config from a json/xml/.properties file. We'll likely need to experiment with a few approaches here, to strike a balance.

uchagani:
Resolves #1369 

@yury-s here is an initial implementation of `Playwright` and `Browser` fixtures.  This implementation works for both `concurrent` and `same_thread` modes.  I will add `BrowserContext` and `Page` implementations after your feedback.

Notes:
- I added comments where needed to explain things but will be removed when reviewed.
- I made changes to the `Browser` API to do some testing.  While I think these would be good to have, they will be removed once reviewed.
- The tests can probably be removed if we decide to use this in the actual playwright-java tests but for now  i needed to see if everything is working so i added these tests


playwright-junit/src/main/java/com/microsoft/playwright/BrowserExtension.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private static BrowserFactory getBrowserFactoryInstance(ExtensionContext extensionContext) {
+    UseBrowserFactory useBrowserFactory =
+      findAnnotation(extensionContext.getTestMethod(), UseBrowserFactory.class)
+        .orElse(findAnnotation(extensionContext.getTestClass(), UseBrowserFactory.class)
+          .orElseThrow(() -> new PlaywrightException("UseBrowserFactory annotation not found.")));
+
+    try {
+      return useBrowserFactory.value().newInstance();
```
yury-s, line 73: We might want to cache `BrowserFactory` instance and create it just once per thread/run etc. For now creating all it every time is fine.

playwright-junit/src/main/java/com/microsoft/playwright/BrowserExtension.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    Playwright playwright = getOrCreatePlaywright();
+    Browser browser = factory.newBrowser(playwright);
+    System.out.println("Created Browser " + browser);
+    threadLocalBrowserMap.get().put(browserFactoryClass, browser);
+    return browser;
+  }
+
+  static void closeAllBrowsers() {
```
yury-s, line 51: Let's close the browser in afterAll callback in this class. Also if you wan to use per method annotation for browser factory, the browser should be closed after such test method finished running (aka afterEach). But I'd keep things simple for now and only support class-level `UseBrowserFactory` annotations, WDYT?

playwright-junit/src/main/java/com/microsoft/playwright/PlaywrightExtension.java
```diff
@@ -0,0 +1,50 @@
...
+      System.out.println("Created Playwright " + getPlaywright());
+    }
+    return getPlaywright();
+  }
+
+  public static Playwright getPlaywright() {
+    return playwrightThreadLocal.get();
+  }
+
+  public static void closePlaywright() {
```
yury-s, line 43: Ideally we'd keep `Playwright` instances until all the tests finished running. Something like [RunListener.testRunFinished](https://junit.org/junit4/javadoc/4.12/org/junit/runner/notification/RunListener.html#testRunFinished(org.junit.runner.Result)) seems a good hook to run such logic but it looks like that interface was removed in JUnit 5. Maybe there is a new counterpart?

playwright-junit/src/main/java/com/microsoft/playwright/ExtensionUtils.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.Locale;
+
+import static org.junit.platform.commons.support.AnnotationSupport.findAnnotation;
+import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
+
+class ExtensionUtils {
+  // Necessary because of https://github.com/junit-team/junit5/issues/3447
+  static ExecutionMode getExecutionMode(ExtensionContext extensionContext) {
+    return findAnnotation(extensionContext.getTestMethod(), Execution.class)
```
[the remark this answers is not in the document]
uchagani: We would still need this method because of the bug I referenced unfortunately. 

Supporting per method is not adding any extra complexity aside from knowing when to close the objects. 

The creation of all objects, regardless of happens when the ParameterResolver runs and that would be up to the user whether they need it in a hook or the test method.

playwright-junit/src/main/java/com/microsoft/playwright/PlaywrightExtension.java
```diff
@@ -0,0 +1,50 @@
...
+    Class<?> clazz = parameterContext.getParameter().getType();
+    return Playwright.class.equals(clazz);
+  }
+
+  @Override
+  public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+    return getOrCreatePlaywright();
+  }
+
+  public static Playwright getOrCreatePlaywright() {
```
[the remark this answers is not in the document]
uchagani: I made these public so users could build their own extensions on top of junit-playwright.

For example, creating auto-navigation extensions or auto-login extensions for their test frameworks.

For now, we can make it private if you'd like but I think this is something that we should allow at some point.

playwright-junit/src/main/java/com/microsoft/playwright/PlaywrightCleanupExtension.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+import static com.microsoft.playwright.BrowserExtension.closeAllBrowsers;
+import static com.microsoft.playwright.ExtensionUtils.getExecutionMode;
+import static com.microsoft.playwright.PlaywrightExtension.closePlaywright;
+
+public class PlaywrightCleanupExtension implements AfterEachCallback, AfterAllCallback {
```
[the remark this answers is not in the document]
uchagani: We can but then the registration of extensions comes into play regarding which AfterAllCallback hook runs first (we would want browser then playwright in that order). Which is okay since we are registering extensions with our custom annotation.

playwright-junit/src/main/java/com/microsoft/playwright/BrowserExtension.java
uchagani: Makes sense, should be doable with another map.

playwright-junit/src/main/java/com/microsoft/playwright/BrowserExtension.java
uchagani: I think we don't don't currently want to support method-level parallelizations then we shouldn't support method-level custom factory. That way we don't have to deal with any after each callbacks

playwright-junit/src/main/java/com/microsoft/playwright/PlaywrightExtension.java
uchagani: Junit 5 doesn't have a built in after test run hook.

They have a workaround using a closeable resource but it currently could run in a separate thread than the class which means we couldn't access the threadlocal objects.

I've opened an issue with JUnit for this also.

For now, I think after class is our best option.

uchagani:
@yury-s removed all method-level parallelization code and made the other updates you suggested.  If you like this approach, I can continue with BrowserContext and Page fixtures.

playwright-junit/src/main/java/com/microsoft/playwright/junit/BrowserExtension.java
```diff
@@ -0,0 +1,90 @@
...
+    threadLocalBrowserMap.remove();
+  }
+
+  private static Browser getBrowser(Class<? extends BrowserFactory> browserFactoryClass) {
+    return threadLocalBrowserMap.get().get(browserFactoryClass);
+  }
+
+  private static BrowserFactory getBrowserFactoryInstance(ExtensionContext extensionContext) {
+    UseBrowserFactory useBrowserFactory = findAnnotation(extensionContext.getTestClass(), UseBrowserFactory.class)
+      .orElseThrow(() -> new PlaywrightException("UseBrowserFactory annotation not found."));
```
yury-s, line 71: How can it throw if `supportsParameter` is called first and checks that the annotation is present?

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserDefaultFactoryTests.java
```diff
@@ -0,0 +1,56 @@
...
+    assert browser != null;
+    System.out.println("BeforeAll " + browser);
+    browserFromBeforeAll = browser;
+  }
+
+  @AfterAll
+  static void afterAll(Browser browser) {
+    assert browser != null;
+    System.out.println("AfterAll " + browser);
+    assertEquals(browserFromBeforeAll, browser, "Static Browser is not equal to others that were created");
```
yury-s, line 24: Is it guaranteed that @afterAll of BrowserExtension is called after all afterAll's in the test class?

playwright-junit/src/main/java/com/microsoft/playwright/junit/BrowserExtension.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.microsoft.playwright.junit.PlaywrightExtension.getOrCreatePlaywright;
+import static org.junit.platform.commons.support.AnnotationSupport.findAnnotation;
+import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
+
+public class BrowserExtension implements ParameterResolver, AfterAllCallback {
+  private static final ThreadLocal<Map<Class<? extends BrowserFactory>, Browser>> threadLocalBrowserMap;
```
yury-s, line 16: Can there be more than one entry in the map? My understanding is that there is at most one active test class and each class has only one UseBrowserFactory annotation.

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserCustomFactoryTests.java
```diff
@@ -0,0 +1,57 @@
+package com.microsoft.playwright.junit;
+
+import com.microsoft.playwright.Browser;
+import org.junit.jupiter.api.*;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@UseBrowserFactory(CustomBrowserFactory.class)
```
yury-s, line 8: I wonder if it'd be more appropriate to make this a parameter/field annotation. It'd be probably more verbose but also more explicit, what are you thoughts?

yury-s (commented):
Left a few comments, but I like this much better. I think we should continue in this direction.

playwright-junit/src/main/java/com/microsoft/playwright/junit/BrowserExtension.java
uchagani: It can't, I just thought but would be better than null but I'll look at this again to find a better way.

playwright-junit/src/main/java/com/microsoft/playwright/junit/BrowserExtension.java
uchagani: Good catch. This is a left over from when we would allow more than one browser per class. I'll fix it.

playwright-junit/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.microsoft.playwright</groupId>
+        <artifactId>parent-pom</artifactId>
+        <version>1.38.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>playwright-junit</artifactId>
```
jfgreffier, line 12: It would be nice to also have a <name> and <description>

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserCustomFactoryTests.java
uchagani: Are you thinking something like this:

```java
@Test
void foo(@UseBrowserFactory(CustomBrowser.class) Page page) {

  //test

}
```

Personally I think class level annotation is good as it is saying the at all tests in this class will use this browser factory.

We can allow further customizing in future PRs.

yury-s: Yes, I meant something like @TempDir junit annotation. I agree that having one class-level annotation is more convenient, let's stick with that, can definitely be customized in the future.

playwright-junit/src/main/java/com/microsoft/playwright/junit/PlaywrightExtension.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Override
+  public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+    return getOrCreatePlaywright();
+  }
+
+  static Playwright getOrCreatePlaywright() {
+    if (playwrightThreadLocal.get() == null) {
+      playwrightThreadLocal.set(Playwright.create());
+      System.out.println("Created Playwright " + getPlaywright());
```
yury-s, line 32: Remove logging here and in other places?

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserCustomFactoryTests.java
```diff
@@ -0,0 +1,57 @@
...
+    assert browser != null;
+    System.out.println("BeforeAll " + browser);
+    browserFromBeforeAll = browser;
+    assertEquals("firefox", browser.browserType().name());
+  }
+
+  @AfterAll
+  static void afterAll(Browser browser) {
+    assert browser != null;
+    System.out.println("AfterAll " + browser);
```
yury-s, line 24: Remove debug logging.

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserCustomFactoryTests.java
```diff
@@ -0,0 +1,57 @@
...
+    assertEquals(browserFromBeforeAll, browser, "Static Browser is not equal to others that were created");
+  }
+
+  @BeforeEach
+  void beforeEach(Browser browser) {
+    assert browser != null;
```
yury-s, line 30: `assertNotNull(browser);` here and in other places so that we use junit assertions in the tests and not rely on java assertions to be enabled?

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserDefaultFactoryTests.java
```diff
@@ -0,0 +1,56 @@
+package com.microsoft.playwright.junit;
+
+import com.microsoft.playwright.Browser;
+import org.junit.jupiter.api.*;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@UseBrowserFactory
+public class BrowserDefaultFactoryTests {
```
yury-s, line 9: As we rely on ThreadLocal, can we also add a test for the fixture running on 2 threads concurrently?

playwright-junit/src/test/java/com/microsoft/playwright/junit/PlaywrightTests.java
```diff
@@ -0,0 +1,59 @@
...
+  @Test
+  void test1(Playwright playwright) {
+    assert playwright != null;
+    System.out.println("Test1" + playwright);
+    assertEquals(p, playwright, "Playwright parameter is not equal to others that were created");
+    assertEquals(playwright, ps, "Playwright parameter is not equal to static Playwright from Before/AfterAll hooks");
+  }
+
+  @Test
+  void test2(Playwright playwright) {
```
yury-s, line 53: What if someone has (intentionally or by accident) 2 Playwright parameters, will it be one and the same instance? Let's add a test for that. Same for the Browser fixture.

yury-s (commented):
> I will add BrowserContext and Page implementations after your feedback.

Do you want to add them to the same PR or follow-up in a separate one?

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserCustomFactoryTests.java
yury-s: One thing that is not very clear with class level annotation is fields with type `Browser`, but it's probably fine if they are not assigned by the fixture.

playwright-junit/src/test/java/com/microsoft/playwright/junit/BrowserCustomFactoryTests.java
uchagani: I was thinking of removing these tests altogether and using playwright-junit in the actual playwright tests.  what do you think?  that way we won't need a separate test suite for this.  If we do want a separate test suite, JUnit provides a way to test extensions that we could use to properly test the extensions.

yury-s: That'd be nice and would also help us evaluate usability of the fixtures.

uchagani: @yury-s I'm running into a circular dependency issue when trying to update the playwright tests to using playwright-junit.

Since playwright-junit needs playwright as a dependency, do you know of a clean way to use playwright-junit in playwright?  

i can think of 2 ways:

1.  include junit-playwright inside of the main playwright package but with the junit dependency as provided scope.  
2. put junit-playwright in a completely separate maven project

What are your thoughts?

yury-s: If I understand correctly, marking `playwright` dependency of `playwright-junit` `provided` should enable using it in the `playwright` tests, is it not working? I believe you need to remove `playwright` from the parent pom.xml in that case.

yury-s: If this does not work, including junit stuff into the main playwright module with "provided" junit dependency is okay too.

uchagani: Yea i don't think it will work.  If i remove playwright from the parent pom, then maven complains it can't find it inside `playwright-junit`.  And once i add `playwright-junit` to `playwright`, i get

```
java: Annotation processing is not supported for module cycles. Please ensure that all modules from cycle [playwright,playwright-junit] are excluded from annotation processing
```

uchagani: I'll just include it in the main playwright module

uchagani:
@yury-s this is ready for another review.  All playwright fixtures have been added.  I also started updating the tests to use the fixtures.  Please let me know if updating the tests in the manner that I did it is okay, and if so, i can continue updating the rest of them.  

The idea behind the tests is to use the `TestBase` only for the server parts of the tests.  all other playwright-related things should be handled by the fixtures.  See my comments in the test classes themselves

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
```diff
@@ -16,20 +16,25 @@
...
 
-public class TestBrowserContextBasic extends TestBase {
+@Tag("fixtures") //temp tag to allow only tests with junit integration to run.  will be removed before merge.
+@UsePlaywright(browserFactory = BrowserFromEnv.class)
+public class TestBrowserContextBasic extends __TestBaseNew {
   @Test
-  void shouldCreateNewContext() {
+  void shouldCreateNewContext(Browser browser) {
+    //Failing: should be zero?
     assertEquals(1, browser.contexts().size());
```
uchagani, line 38: @yury-s this and another test below are failing for the same reason.  If using fixtures, i think these assertions should be updated to be 0 for the context list size.

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
```diff
@@ -92,11 +97,12 @@ void shouldIsolateLocalStorageAndCookies() {
     context1.close();
     context2.close();
 
+    //Failing is 1 expected because of the old TestBase class? Should it be zero since we are only creating 2 contexts and closing both?
     assertEquals(1, browser.contexts().size());
```
uchagani, line 101: here is the other place it is failing because of the old TestBase's default browser context creation.

yury-s: Yes, you are right. Previously we were creating a context unconditionally in beforeEach and we assert that the closed contexts are removed from the contexts() list.

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
yury-s: Yes, updating to 0 sounds right, see my response below.

yury-s (commented):
> Please let me know if updating the tests in the manner that I did it is okay, and if so, i can continue updating the rest of them.

Updated tests look good, but since it is going to be a lot of mostly mechanical changes in many files I suggest we do it in a separate PR and keep this one focused on the implementation. Having just a few tests that you had in the previous iteration should be fine for the initial implementation and should unblock this PR.

playwright/src/main/java/com/microsoft/playwright/junit/BrowserExtension.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.microsoft.playwright.Browser;
+import com.microsoft.playwright.Playwright;
+import com.microsoft.playwright.PlaywrightException;
+import org.junit.jupiter.api.extension.*;
+
+import static com.microsoft.playwright.junit.ExtensionUtils.getUsePlaywrightAnnotation;
+import static com.microsoft.playwright.junit.ExtensionUtils.hasUsePlaywrightAnnotation;
+
+class BrowserExtension implements ParameterResolver, AfterAllCallback {
```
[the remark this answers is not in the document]
uchagani: If we move the files into another package we will need to make the extensions public instead of package-private.  Do we want to expose them?

playwright/src/main/java/com/microsoft/playwright/junit/DefaultPlaywrightFactory.java
```diff
@@ -0,0 +1,10 @@
+package com.microsoft.playwright.junit;
+
+import com.microsoft.playwright.Playwright;
+
+class DefaultPlaywrightFactory implements PlaywrightFactory {
```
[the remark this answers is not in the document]
uchagani: They don't have to be public since they are only used internally.  but we can make them public if you'd like

playwright/src/test/java/com/microsoft/playwright/BrowserFromEnv.java
```diff
@@ -0,0 +1,34 @@
+package com.microsoft.playwright;
+
+import com.microsoft.playwright.junit.BrowserFactory;
+
+import static com.microsoft.playwright.Utils.getBrowserChannelFromEnv;
+import static com.microsoft.playwright.Utils.getBrowserNameFromEnv;
+
+public class BrowserFromEnv implements BrowserFactory {
```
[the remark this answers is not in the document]
uchagani: Do we want to use `HEADLESS` instead so that the API to configure tests and for playwright are the same?

uchagani:
@yury-s left some more comments, for some reason my comments before were showing up as pending and i had to submit the review.

playwright/src/test/java/com/microsoft/playwright/BrowserFromEnv.java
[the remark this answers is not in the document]
yury-s: `HEADED` is preferred as it defaults to `false` when not specified (rather than headless that is true by default and only if specified as HEADLESS=0 it actually changes the behavior), consistent with --headed in node.js command line and  `HEADED` in .Net. We didn't change the parameter in the library API as it had already been shipped by the time we realized this in the test runner.

playwright/src/main/java/com/microsoft/playwright/junit/DefaultPlaywrightFactory.java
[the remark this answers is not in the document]
yury-s: > They don't have to be public since they are only used internally. 

Can people use use the explicitly in the annotations today (`@UsePlaywright(playwright = DefaultPlaywrightFactory.class)`) ?

> but we can make them public if you'd like

I see 2 reasons to make them public: people may want to subclass them, they may want to use them explicitly in the annotations.

playwright/src/main/java/com/microsoft/playwright/junit/BrowserExtension.java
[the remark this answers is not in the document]
yury-s: Yeah, I think it is fine, although not ideal. There are already precedents of this, e.g. PlaywrightImpl. We don't make any promises about backwards compatibility of the classes in the `impl` package, those who use them are on their own.

yury-s:
> @yury-s left some more comments, for some reason my comments before were showing up as pending and i had to submit the review.

Yeah, that's sometimes annoying that they are not published immediately.

playwright/src/main/java/com/microsoft/playwright/junit/DefaultPlaywrightFactory.java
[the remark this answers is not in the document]
uchagani: Sure I can make them public. I figured no one would explicitly use the default ones and would define their own if they wanted something custom.

uchagani:
@yury-s addressed all previous suggestions.  ready for another review.

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
```diff
@@ -0,0 +1,77 @@
...
+import static com.microsoft.playwright.junit.impl.ExtensionUtils.getUsePlaywrightAnnotation;
+import static com.microsoft.playwright.junit.impl.ExtensionUtils.hasUsePlaywrightAnnotation;
+
+public class BrowserExtension implements ParameterResolver, AfterAllCallback {
+  private static final ThreadLocal<Browser> threadLocalBrowser;
+  private static final ThreadLocal<BrowserFactory> threadLocalBrowserFactory;
+
+  static {
+    threadLocalBrowser = new ThreadLocal<>();
+    threadLocalBrowserFactory = new ThreadLocal<>();
```
yury-s, line 19: ditto

yury-s (approved):

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserContextExtension.java
```diff
@@ -0,0 +1,87 @@
...
+        browserContext.close();
+      }
+    } finally {
+      threadLocalBrowserContext.remove();
+    }
+  }
+
+  private void cleanupBrowserContextFactory() {
+    BrowserContextFactory browserContextFactory = threadLocalBrowserContextFactory.get();
+    if (browserContextFactory != null) {
```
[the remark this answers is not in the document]
uchagani: If we drop it we would have to recreate the factory for each test. The way we currently have it, we only need to create the factory the first time and reuse it for the entire test class.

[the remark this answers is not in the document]
yury-s: I thought it was already happening when `cleanupBrowserContextFactory` is called from `afterEach`, basically the thread local cache is cleared between the test runs. Am I missing something?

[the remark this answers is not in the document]
uchagani: That is a good catch, it should not be cleaned up in the `afterEach` but in an `afterAll`

playwright/src/test/java/com/microsoft/playwright/TestFixturesWithCustomFactories.java
```diff
@@ -0,0 +1,93 @@
...
+    playwrightFromBeforeAll = playwright;
+    browserFromBeforeAll = browser;
+  }
+
+  @AfterAll
+  static void afterAll(Playwright playwright, Browser browser) {
+    assertNotNull(playwright);
+    assertNotNull(browser);
+    assertEquals(playwrightFromBeforeAll, playwright);
+    assertEquals(browserFromBeforeAll, browser);
```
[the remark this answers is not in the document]
uchagani: i'm not sure how we can do this without using JUnit's extension testing engine.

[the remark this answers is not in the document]
yury-s: Ok, never mind then. I thought about something simple like a static/thread-local counter field inside CustomBrowserFactory and a check in afterAll of the test. We can certainly live without it.

uchagani:
@yury-s made the latest changes:  

* `BrowserContextFactory` is now cleaned up in `AfterAllCallback` instead of `AfterEachCallback`
* invalid `BROWSER` value will now throw.

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserContextExtension.java
```diff
@@ -0,0 +1,91 @@
...
```
yury-s, line 87: nit: `threadLocalBrowserContextFactory.remove();` can be called unconditionally as it is noop when there is no value.

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserContextExtension.java
```diff
@@ -0,0 +1,91 @@
...
+      threadLocalBrowserContextFactory.set(browserContextFactory);
+      return browserContextFactory;
+    } catch (InstantiationException | IllegalAccessException e) {
+      throw new PlaywrightException("Unable to create an instance of the supplied BrowserContextFactory", e);
+    }
+  }
+
+  private void cleanupBrowserContext() {
+    try {
+      BrowserContext browserContext = threadLocalBrowserContext.get();
```
yury-s, line 76: nit: maybe restructure it like this to avoid try/finally?

```java
      BrowserContext browserContext = threadLocalBrowserContext.get();
      threadLocalBrowserContext.remove();
      if (browserContext != null) {
        browserContext.close();
      }
```

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
```diff
@@ -0,0 +1,77 @@
...
+    Playwright playwright = PlaywrightExtension.getOrCreatePlaywright(extensionContext);
+    BrowserFactory browserFactory = getBrowserFactoryInstance(extensionContext);
+    browser = browserFactory.newBrowser(playwright);
+    threadLocalBrowser.set(browser);
+    return browser;
+  }
+
+  private static BrowserFactory getBrowserFactoryInstance(ExtensionContext extensionContext) {
+    if (threadLocalBrowserFactory.get() != null) {
+      return threadLocalBrowserFactory.get();
```
yury-s, line 64: nit: this method is called just once per class when Browser instance is being created (and cached), there is not much point in caching the factory I believe.

yury-s (approved):
A few nits, but I'm okay with merging in its current form.

uchagani:
@yury-s can I get another approval. Had to fix a test.

yury-s (approved):

yury-s:
`TestFixturesWithDefaultFactories.beforeAll:20 expected: <null> but was: <chromium>` for some reason is failing in Docker, the other failures on the bots look unrelated.

uchagani:
Could be in docker we aren't setting `BROWSER`. I can update the test to account for that.

yury-s:
> Could be in docker we aren't setting `BROWSER`. I can update the test to account for that.

We don't.

uchagani:
@yury-s, fixed.

yury-s (approved):

PR #1376, opened 2023-09-06
chore: support js Map, Set in protocol results

yury-s:
Upstream PR: https://github.com/microsoft/playwright/pull/26730
Reference: https://github.com/microsoft/playwright-java/issues/1353

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java
@@ -281,15 +281,10 @@ class Serialization {
       return (T) map;
     }
     if (value.m != null) {
-      List<List<Object>> entries = deserialize(value.m);
-      LinkedHashMap map = new LinkedHashMap();
-      for (List<Object> entry: entries) {
-        map.put(entry.get(0), entry.get(1));
-      }
-      return (T) map;
+      return (T) new LinkedHashMap();
     }
     if (value.se != null) {
-      return (T) deserialize(value.se);
+      return (T) new LinkedHashMap();
     }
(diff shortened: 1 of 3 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/Serialization.java
```diff
@@ -280,6 +280,12 @@ private static <T> T deserialize(SerializedValue value, Map<Integer, Object> idT
       }
       return (T) map;
     }
+    if (value.m != null) {
+      return (T) new LinkedHashMap();
```
dgozman, line 284: Should we do `idToValue.put(value.id, map)` here as well?

dgozman (approved):

playwright/src/test/java/com/microsoft/playwright/TestPageEvaluate.java
```diff
@@ -645,4 +645,14 @@ void shouldAcceptParameter() {
     assertTrue(object instanceof Date);
     assertEquals(Date.from(instant), object);
   }
+
+  @Test
+  void shouldTransferMaps() {
+    assertEquals(mapOf(), page.evaluate("() => new Map([[1, { test: 42n }]])"));
```
YusukeIwaki, line 651: @yury-s Hi, does this mean playwright-java doesn't support transfering Map of Set??

YusukeIwaki: Solved with this comment https://github.com/microsoft/playwright-python/pull/2075#discussion_r1328099084,
thank you.

PR #1412, opened 2023-10-13
feat(junit-playwright)

uchagani:
@yury-s, new branch/PR with updated options as per your comments in #1401

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -4,15 +4,14 @@ import com.microsoft.playwright.APIRequest;
 import com.microsoft.playwright.Browser;
 import com.microsoft.playwright.BrowserType;
 import com.microsoft.playwright.Playwright;
+import com.microsoft.playwright.options.ViewportSize;
 
 import java.nio.file.Path;
-import java.util.List;
-import java.util.Map;
 
 public class Options {
   private String baseUrl;
   private Path storageStatePath;
-  private Viewport viewport;
+  private ViewportSize viewportSize;
   private String channel;
   private boolean headless;
   private String browserName = "chromium";
@@ -75,15 +74,6 @@ public class Options {
     return this;
   }
 
-  public Viewport getViewport() {
-    return viewport;
-  }
-
-  public Options setViewport(Viewport viewport) {
-    this.viewport = viewport;
-    return this;
-  }
-
   public String getBrowserName() {
     return browserName;
   }
@@ -111,27 +101,12 @@ public class Options {
     return this;
   }
 
-  public static class Viewport {
-    private String width;
-    private String height;
-
-    public String getWidth() {
-      return width;
-    }
-
-    public Viewport setWidth(String width) {
-      this.width = width;
-      return this;
-    }
-
-    public String getHeight() {
-      return height;
-    }
-
-    public Viewport setHeight(String height) {
-      this.height = height;
-      return this;
-    }
+  public ViewportSize getViewportSize() {
+    return viewportSize;
   }
 
+  public Options setViewportSize(ViewportSize viewportSize) {
+    this.viewportSize = viewportSize;
+    return this;
+  }
 }
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -15,6 +15,9 @@ public class Options {
   private String channel;
   private boolean headless;
   private String browserName;
+  private BrowserType.LaunchOptions launchOptions;
+  private Browser.NewContextOptions contextOption;
+  private APIRequest.NewContextOptions apiRequestOptions;
 
   public BrowserType.LaunchOptions getLaunchOptions() {
     return launchOptions;
@@ -42,10 +45,6 @@ public class Options {
     this.apiRequestOptions = apiRequestOptions;
     return this;
   }
-
-  private BrowserType.LaunchOptions launchOptions;
-  private Browser.NewContextOptions contextOption;
-  private APIRequest.NewContextOptions apiRequestOptions;
 
   public String getBaseUrl() {
     return baseUrl;
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -9,16 +9,16 @@ import com.microsoft.playwright.options.ViewportSize;
 import java.nio.file.Path;
 
 public class Options {
-  private String baseUrl;
-  private Path storageStatePath;
-  private ViewportSize viewportSize;
-  private String channel;
-  private Boolean headless;
-  private String browserName = "chromium";
-  private BrowserType.LaunchOptions launchOptions;
-  private Browser.NewContextOptions contextOption;
-  private APIRequest.NewContextOptions apiRequestOptions;
-  private Playwright.CreateOptions playwrightCreateOptions;
+  public String baseUrl;
+  public Path storageStatePath;
+  public ViewportSize viewportSize;
+  public String channel;
+  public Boolean headless;
+  public String browserName = "chromium";
+  public BrowserType.LaunchOptions launchOptions;
+  public Browser.NewContextOptions contextOption;
+  public APIRequest.NewContextOptions apiRequestOptions;
+  public Playwright.CreateOptions playwrightCreateOptions;
 
   public Playwright.CreateOptions getPlaywrightCreateOptions() {
     return playwrightCreateOptions;
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -3,6 +3,7 @@ package com.microsoft.playwright.junit;
 import com.microsoft.playwright.APIRequest;
 import com.microsoft.playwright.Browser;
 import com.microsoft.playwright.BrowserType;
+import com.microsoft.playwright.Playwright;
 
 import java.nio.file.Path;
 import java.util.List;
@@ -14,10 +15,20 @@ public class Options {
   private Viewport viewport;
   private String channel;
   private boolean headless;
-  private String browserName;
+  private String browserName = "chromium";
   private BrowserType.LaunchOptions launchOptions;
   private Browser.NewContextOptions contextOption;
   private APIRequest.NewContextOptions apiRequestOptions;
+  private Playwright.CreateOptions playwrightCreateOptions;
+
+  public Playwright.CreateOptions getPlaywrightCreateOptions() {
+    return playwrightCreateOptions;
+  }
+
+  public Options setPlaywrightCreateOptions(Playwright.CreateOptions playwrightCreateOptions) {
+    this.playwrightCreateOptions = playwrightCreateOptions;
+    return this;
+  }
 
   public BrowserType.LaunchOptions getLaunchOptions() {
     return launchOptions;
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -10,7 +10,7 @@ import java.util.Map;
 
 public class Options {
   private String baseUrl;
-  private Path storageState;
+  private Path storageStatePath;
   private Viewport viewport;
   private String channel;
   private boolean headless;
@@ -55,12 +55,12 @@ public class Options {
     return this;
   }
 
-  public Path getStorageState() {
-    return storageState;
+  public Path getStorageStatePath() {
+    return storageStatePath;
   }
 
-  public Options setStorageState(Path storageState) {
-    this.storageState = storageState;
+  public Options setStorageStatePath(Path storageStatePath) {
+    this.storageStatePath = storageStatePath;
     return this;
   }
 
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -13,7 +13,7 @@ public class Options {
   private Path storageStatePath;
   private ViewportSize viewportSize;
   private String channel;
-  private boolean headless;
+  private Boolean headless;
   private String browserName = "chromium";
   private BrowserType.LaunchOptions launchOptions;
   private Browser.NewContextOptions contextOption;
@@ -92,11 +92,11 @@ public class Options {
     return this;
   }
 
-  public boolean isHeadless() {
+  public Boolean isHeadless() {
     return headless;
   }
 
-  public Options setHeadless(boolean headless) {
+  public Options setHeadless(Boolean headless) {
     this.headless = headless;
     return this;
   }
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -1,0 +1,127 @@
+package com.microsoft.playwright.junit;
+
+import com.microsoft.playwright.APIRequest;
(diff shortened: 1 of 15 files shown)


uchagani:
@yury-s once the options are approved I'll start the implementation to use the options.

yury-s:
> @yury-s once the options are approved I'll start the implementation to use the options.

This list of options looks good to me, we can start with it.

playwright/src/main/java/com/microsoft/playwright/junit/Options.java
```diff
@@ -0,0 +1,126 @@
...
+  public Options setBaseUrl(String baseUrl) {
+    this.baseUrl = baseUrl;
+    return this;
+  }
+
+  public Path getStorageState() {
+    return storageState;
+  }
+
+  public Options setStorageState(Path storageState) {
```
yury-s, line 62: nit: it's named setStorageStatePath in the context options, so it may make sense to stay consistent with that.

uchagani:
@yury-s i haven't forgotten about this.  Life came up.  Will work on it this weekend.

uchagani:
@yury-s, as I was implementing the extensions with options, I came across the following gap:

Now that we are using config options to configure the objects instead of allowing the user to return the objects they want to use, how did you want to handle situations like `BrowserType.connect` or `BrowserType.connectOverCDP`?  Before the user could do whatever they wanted in the `BrowserFactory` as long as they returned a `Browser` but now that is not possible.

yury-s:
> @yury-s, as I was implementing the extensions with options, I came across the following gap:
> 
> Now that we are using config options to configure the objects instead of allowing the user to return the objects they want to use, how did you want to handle situations like `BrowserType.connect` or `BrowserType.connectOverCDP`? Before the user could do whatever they wanted in the `BrowserFactory` as long as they returned a `Browser` but now that is not possible.

Sorry for late reply, was distracted by other stuff too.  We can add connect options later and make the take precedence over launch options (if any), i.e. if there are connect options we'll connect to a remote browser instead of launching one locally. This part is quite a mess in Node.js too, so I'd leave it for later and focus on local browsers instead. Does it make sense?

uchagani:
@yury-s this is ready for another review.  All extensions have been implemented.  Another question: do we want to expose `APIRequestContext` as a fixture?

playwright/src/main/java/com/microsoft/playwright/junit/impl/OptionsExtension.java
```diff
@@ -0,0 +1,36 @@
...
+
+  static Options getOptions(ExtensionContext extensionContext) {
+    Options options = threadLocalOptions.get();
+    if (options != null) {
+      return options;
+    }
+
+    UsePlaywright usePlaywrightAnnotation = getUsePlaywrightAnnotation(extensionContext);
+    try {
+      OptionsFactory configFactory = usePlaywrightAnnotation.optionsFactory().newInstance();
```
yury-s, line 28: Perhaps we can avoid this indirection and allow passing options sub class as a parameter instead, so that it can be instantiated here? It'd require extending `Options` class every time instead of just setting properties on one and the same type but would eliminate the indirection added by the factory. It seems to be roughly the same amount of code on the client end with both approaches so I don't have strong preference. Any thoughts on this?

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
```diff
@@ -0,0 +1,81 @@
...
+      default:
+        throw new PlaywrightException("Invalid browser name.");
+    }
+
+    threadLocalBrowser.set(browser);
+    return browser;
+  }
+
+  private static BrowserType.LaunchOptions getLaunchOptions(Options options) {
+    BrowserType.LaunchOptions launchOptions = options.getLaunchOptions();
```
yury-s, line 68: Let's [clone](https://github.com/microsoft/playwright-java/blob/4efb792c36ea89d88638993076576c33519f3413/playwright/src/main/java/com/microsoft/playwright/impl/Utils.java#L84) the options object to not modify original instance when applying channel and headless overrides below.

playwright/src/main/java/com/microsoft/playwright/junit/Options.java
```diff
@@ -0,0 +1,112 @@
...
+import com.microsoft.playwright.options.ViewportSize;
+
+import java.nio.file.Path;
+
+public class Options {
+  private String baseUrl;
+  private Path storageStatePath;
+  private ViewportSize viewportSize;
+  private String channel;
+  private boolean headless;
```
yury-s, line 16: Let's make it `Boolean` to make it clear whe `LaunchOptions.headless` should not be overridden (similar to how you check `channel`).

playwright/src/main/java/com/microsoft/playwright/junit/impl/PageExtension.java
```diff
@@ -0,0 +1,46 @@
...
+  public void afterEach(ExtensionContext extensionContext) {
+    Page page = threadLocalPage.get();
+    threadLocalPage.remove();
+    if(page != null) {
+      page.close();
+    }
+  }
+
+  @Override
+  public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
```
yury-s, line 22: `supportsParameter/resolveParameter` logic is very similar across the extensions, consider extracting this in a common superclass parametrized by the parameter class.

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserContextExtension.java
```diff
@@ -0,0 +1,71 @@
...
+    Options options = OptionsExtension.getOptions(extensionContext);
+    Browser browser = BrowserExtension.getOrCreateBrowser(extensionContext);
+    Browser.NewContextOptions contextOptions = getContextOptions(options);
+    browserContext = browser.newContext(contextOptions);
+    threadLocalBrowserContext.set(browserContext);
+    return browserContext;
+  }
+
+  private static Browser.NewContextOptions getContextOptions(Options options) {
+    Browser.NewContextOptions contextOptions = options.getContextOption();
```
yury-s, line 52: Let's also make a [copy](https://github.com/microsoft/playwright-java/blob/4efb792c36ea89d88638993076576c33519f3413/playwright/src/main/java/com/microsoft/playwright/impl/Utils.java#L84) before overriding options.

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
```diff
@@ -0,0 +1,81 @@
...
+    return Browser.class.equals(clazz);
+  }
+
+  @Override
+  public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+    return getOrCreateBrowser(extensionContext);
+  }
+
+  static Browser getOrCreateBrowser(ExtensionContext extensionContext) {
+    Browser browser = threadLocalBrowser.get();
```
yury-s, line 39: This is good as the first step. `Browser` and `Playwright` instances could be reused by tests that not run concurrently (to avoid unsafe multithreaded access) if we want. We do this in .net port already. Not necessary for this PR, just a thought for the future iterations.

yury-s (approved):
> Another question: do we want to expose APIRequestContext as a fixture?

Since we have it in the `Options` we should provide the fixture too. It proved convenient in node.js and creating it doesn't cost anything at runtime, so I'd have it for feature parity.

pom.xml
```diff
@@ -86,6 +86,17 @@
         <version>${websocket.version}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>com.microsoft.playwright</groupId>
+        <artifactId>playwright</artifactId>
```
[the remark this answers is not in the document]
uchagani: Good catch, not needed anymore since we're putting it in the same project.

playwright/src/main/java/com/microsoft/playwright/junit/impl/OptionsExtension.java
uchagani: makes sense, fixed.

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
uchagani: @yury-s are you okay with making the Utils class and this clone method public so I can use it here?

playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
uchagani: yea let's think about this.  this will require us keep track of the currently used playwrights and the threads that they're running on (I think).  something for the future.

yury-s:
@uchagani are we just waiting for the last bit about copying options or you want to merge the PR in its current form?

uchagani:
@yury-s just waiting on the last part and the APIRequestContext fixture. I should be able to do it next week after the holidays

uchagani:
@yury-s, in order to use the `clone` method, we will need to make the fields in `Options` public or add the `setAccessible(true)` for both `fromField` and `toField` inside the `convertType` method.  Which one do you prefer?

yury-s:
> @yury-s, in order to use the `clone` method, we will need to make the fields in `Options` public or add `setAccessible(true)` for both `fromField` and `toField` inside the `convertType` method. Which one do you prefer? For now, I just made the fields in the `Options` class public.
> 

I'd just make them public. So far it worked good for other *Options classes.

> Another question: Currently, we don't allow `BrowserContext` and `Page` parameters to resolve in class-level hooks (`BeforeAll/AfterAll`). Do we want to do the same for `APIRequestContext`? For now I am not allowing them in class level hooks to follow the `BrowserContext` pattern. Let me know what you think.

In Node.js `request` fixture is available in beforeAll/afterAll hooks. We can make it available in `BeforeAll/AfterAll` in Java too for consistency. No strong opinion though, it didn't cause anywhere as much confusion as different page/context instances in the tests and beforeAll hooks in Node.

yury-s (approved):

uchagani:
@yury-s made APIRequestContext available in BeforeAll/AfterAll hooks.  This PR is ready.

playwright/src/main/java/com/microsoft/playwright/junit/impl/APIRequestContextExtension.java
```diff
@@ -0,0 +1,40 @@
...
+import com.microsoft.playwright.junit.Options;
+import org.junit.jupiter.api.extension.*;
+
+import static com.microsoft.playwright.junit.impl.ExtensionUtils.isParameterSupported;
+
+public class APIRequestContextExtension implements ParameterResolver, AfterAllCallback {
+  private static final ThreadLocal<APIRequestContext> threadLocalAPIRequestContext = new ThreadLocal<>();
+
+  @Override
+  public void afterAll(ExtensionContext extensionContext) {
```
yury-s, line 14: This would mean that we use the same APIRequestContext across all tests in the file and we want to create a fresh one for each test (similar to BrowserContext). So we should either find a way to reset the context after all beforeAll hooks have finished and before the tests started or just not have APIRequestContext fixture in beforeAll hooks. Same applies to afterAll hooks. So basically beforeAll should have their own instance of APIRequestContext, each test should have its own and beforeAll hooks should have another instance.

playwright/src/main/java/com/microsoft/playwright/junit/impl/APIRequestContextExtension.java
```diff
@@ -7,9 +7,14 @@
 
 import static com.microsoft.playwright.junit.impl.ExtensionUtils.isParameterSupported;
 
-public class APIRequestContextExtension implements ParameterResolver, AfterAllCallback {
+public class APIRequestContextExtension implements ParameterResolver, BeforeEachCallback, AfterAllCallback {
   private static final ThreadLocal<APIRequestContext> threadLocalAPIRequestContext = new ThreadLocal<>();
 
+  @Override
+  public void beforeEach(ExtensionContext extensionContext) {
```
yury-s, line 14: I'd not do this and keep it in `afterEach` only. Otherwise we might introduce a fixture that uses APIRequestContext (in its beforeEach) which is later removed and the test may end up using another one, different from the beforeEach fixture. Alternatively, we could store APIRequestContext created for beforeAll in a separate thread local and clear only that one in beforeEach and have the test's one cleared in afteEach/

It looks like it could be easier to not provide APIRequestContext in `beforeAll` for now. WDYT?

playwright/src/main/java/com/microsoft/playwright/junit/impl/APIRequestContextExtension.java
```diff
@@ -7,9 +7,14 @@
 
 import static com.microsoft.playwright.junit.impl.ExtensionUtils.isParameterSupported;
 
-public class APIRequestContextExtension implements ParameterResolver, AfterAllCallback {
+public class APIRequestContextExtension implements ParameterResolver, BeforeEachCallback, AfterAllCallback {
```
yury-s, line 10: I believe it should be `BeforeTestExecutionCallback` to avoid the problem described below.

playwright/src/main/java/com/microsoft/playwright/junit/impl/APIRequestContextExtension.java
uchagani: this beforeEach is execute before any user-defined beforeEach, so it's not possible for this beforeEach to remove before the test is executed.  Unless i'm misunderstanding your concern here.

playwright/src/test/java/com/microsoft/playwright/TestPlaywrightFixtures.java
```diff
@@ -0,0 +1,55 @@
...
+  public void objectShouldBeSameAsBeforeAll(Playwright playwright, Browser browser) {
+    assertEquals(playwrightFromBeforeAll, playwright);
+    assertEquals(browserFromBeforeAll, browser);
+  }
+
+  @Test
+  public void objectShouldBeSameAsBeforeEach(BrowserContext browserContext, Page page, APIRequestContext apiRequestContext) {
+    assertEquals(browserContextFromBeforeEach, browserContext);
+    assertEquals(pageFromBeforeEach, page);
+    assertEquals(apiRequestContextFromBeforeEach, apiRequestContext);
```
uchagani, line 53: @yury-s you can see here that I'm verifying that the APIRequestContext is the same as the one from the user-defined BeforeEach hook.

playwright/src/main/java/com/microsoft/playwright/junit/impl/APIRequestContextExtension.java
yury-s: Yes, you are right.

playwright/src/main/java/com/microsoft/playwright/junit/impl/APIRequestContextExtension.java
yury-s: You understood it correctly. Thanks for explaining, let's try it then.

playwright/src/main/java/com/microsoft/playwright/junit/Options.java
```diff
@@ -0,0 +1,112 @@
...
+  public BrowserType.LaunchOptions launchOptions;
+  public Browser.NewContextOptions contextOption;
+  public APIRequest.NewContextOptions apiRequestOptions;
+  public Playwright.CreateOptions playwrightCreateOptions;
+
+  public Playwright.CreateOptions getPlaywrightCreateOptions() {
+    return playwrightCreateOptions;
+  }
+
+  public Options setPlaywrightCreateOptions(Playwright.CreateOptions playwrightCreateOptions) {
```
uchagani, line 27: @yury-s should we remove all the setters from the Options class?  I don't see them being useful since the user would only override the getters, correct?

PR #1425, opened 2023-11-17
Update LocatorImpl.java

cjayswal:
allowing access outside package so that other frameworks can extend and customize.

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/LocatorImpl.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/LocatorImpl.java
@@ -19,7 +19,7 @@ import static com.microsoft.playwright.impl.Serialization.gson;
 import static com.microsoft.playwright.impl.Utils.convertType;
 import static com.microsoft.playwright.impl.Utils.toJsRegexFlags;
 
-class LocatorImpl implements Locator {
+public class LocatorImpl implements Locator {
   final FrameImpl frame;
   final String selector;
 


playwright/src/main/java/com/microsoft/playwright/impl/LocatorImpl.java
```diff
@@ -19,7 +19,7 @@
 import static com.microsoft.playwright.impl.Utils.convertType;
 import static com.microsoft.playwright.impl.Utils.toJsRegexFlags;
 
-class LocatorImpl implements Locator {
+public class LocatorImpl implements Locator {
```
yury-s, line 22: This class is an implementation detail and is not intended to be extended by third parties. Can you start with describing what are you trying to achieve?

cjayswal: I am adding support of playwright to [qaf](https://qmetry.github.io/qaf) - well accepted framework for [selenium](https://www.selenium.dev/ecosystem/#frameworks). It's required to extend original implementation, In order to support features like [custom component](https://qmetry.github.io/qaf/latest/custom_component.html), [locator repository](https://qmetry.github.io/qaf/latest/locator_repository.html), self-descriptive locator.

yury-s: It's not clear from the docs you linked why the features described there would "require to extend" the implementation rather than just using it as a client. We expect all clients to only depend on the public API and use aggregation rather than inheriting from the implementation. Even if you extended LocatorImpl, it would not be interoperable with the rest of the framework (e.g. Page.locator would still return builtin locator). I suggest you start with an issue describing the problem and why it cannot be addressed with current API.

uchagani:
@cjayswal i'm interested in this and have used various design patterns to accomplish this without changing the scope of `LocatorImpl`.  I'm currently implementing custom components in [Stagehand](https://github.com/uchagani/stagehand).  Let me know if you'd like to collaborate.

mxschmitt:
Closing as per https://github.com/microsoft/playwright-java/pull/1425/files#r1399615978.

PR #1436, opened 2023-12-05
Remove setters from Options.  Add tests for custom fixtures

uchagani:
@yury-s I added some tests because i found a bug in the options.  Also removed all setters as they don't seem useful (correct me if i'm wrong)

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestPlaywrightCustomFixtures.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestPlaywrightCustomFixtures.java
@@ -1,0 +1,36 @@
+package com.microsoft.playwright;
+
+import com.microsoft.playwright.junit.Options;
+import com.microsoft.playwright.junit.UsePlaywright;
+import org.junit.jupiter.api.Test;
+
+import java.util.regex.Pattern;
+
+import static com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@UsePlaywright(options = TestPlaywrightCustomFixtures.CustomOptions.class)
+public class TestPlaywrightCustomFixtures {
+  public static class CustomOptions extends Options {
+    @Override
+    public String getBaseUrl() {
+      return "https://bing.com";
+    }
+
+    @Override
+    public String getBrowserName() {
+      return "firefox";
+    }
+  }
+
+  @Test
+  public void testCustomBrowser(Browser browser) {
+    assertEquals(browser.browserType().name(), "firefox");
+  }
+
+  @Test
+  public void testBaseUrl(Page page) {
+    page.navigate("/");
+    assertThat(page).hasURL(Pattern.compile("bing"));
+  }
+}
--- a/playwright/src/test/java/com/microsoft/playwright/TestPlaywrightCustomFixtures.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestPlaywrightCustomFixtures.java
@@ -3,19 +3,38 @@ package com.microsoft.playwright;
 import com.microsoft.playwright.junit.Options;
 import com.microsoft.playwright.junit.OptionsFactory;
 import com.microsoft.playwright.junit.UsePlaywright;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
+import java.io.IOException;
 import java.util.regex.Pattern;
 
+import static com.microsoft.playwright.Utils.nextFreePort;
 import static com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @UsePlaywright(TestPlaywrightCustomFixtures.CustomOptions.class)
 public class TestPlaywrightCustomFixtures {
+  private static Server server;
+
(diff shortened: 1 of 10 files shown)


yury-s:
> @yury-s I added some tests because i found a bug in the options. Also removed all setters as they don't seem useful (correct me if i'm wrong)

The idea with setters is that they could be chained one after another (.setChannel().setHeadless().setViewportSize() etc.). I think we should keep them for convenience and I'd rather remove getters to avoid any confusion, I don't think we want clients to extend options with any smart logic e.g. in `getChannel`, if such logic exists it should compute the channel name and set it once. The extension can read Options field values and be sure that there are no side effects. This is the approach we follow in options in the library APIs.

uchagani:
@yury-s how would setters work when the user has to pass in a class to the `UsePlaywright` annotation?  I think something may have gotten lost in my understanding when we moved away from factories.

yury-s:
> @yury-s how would setters work when the user has to pass in a class to the `UsePlaywright` annotation? I think something may have gotten lost in my understanding when we moved away from factories.

That's a good point. I was thinking about something like:

```java
    new Options().setHeadless(false).setBrowserName("firefox").setViewportSize(1024, 768);
```

but since we need a custom class for the options, it would be:

```java
class MyOptions extends Options {
  public MyOptions() {
    setHeadless(false).setBrowserName("firefox").setViewportSize(1024, 768);
  }
}
```
which is to my taste no better than:

```java
class MyOptions extends Options {
  public MyOptions() {
    headless = false;
    browserName = "firefox";
    viewportSize = new ViewportSize(1024, 768);
  }
}
```
I guess the setters make sense if we allow to specify options factory rather than custom Options class, i.e. something like:

```java
@UsePlaywright(optionsFactory="com.microsoft.test.OptionsFactory#createHeadedFirefoxOptions")
```
this seems a bit too verbose to my taste but perhaps it's better than requiring users to inherit from Options every time. What do you think about these approaches?

uchagani:
@yury-s, I think it might be better if we went back to an OptionsFactory and did something like 

```java
public interface OptionsFactory {
  Options getOptions();
}
```

```java
public class MyOptions implements OptionsFactory { 
  @Override
  public Options getOptions() {
    return new Options().setHeadless(false).setBaseUrl("https://microsoft.com");
  }
}

```java
@UsePlaywright(MyOptions.class)
public class TestClass {
}
```

I think the interface clearly explains what is necessary for the user to provide.  Where as extending the Options class leaves a lot to interpretation and documentation.

yury-s:
> I think the interface clearly explains what is necessary for the user to provide.

Agreed. Leet's start with `@UsePlaywright(optionsFactory= MyOptionsFactory.class)` then? We'll be able to add other ways to parametrize it in the future if we need to (e.g. passing custom options subclass).

We discussed this in the team meeting today and figured that since we don't have strong opinion at this point, we should start with something that would be easier to maintain and change (if needed) in the future based on the client feedback. From that perspective I think it'd be the safest to make all options fields private and have public getters/setters for those that we want to expose to the users. This way we'll have more flexibility to change implementation without breaking existing code. Does it make sense?

uchagani:
@yury-s that makes sense to me. However, in order to use the existing clone method, the fields have to be public. Or I can add the appropriate setAccessible(true) calls in the clone method and make the fields private. What do you prefer?

yury-s:
I'm worried that calling `setAccessible(true)` may cause exceptions is non-standard security configurations. We can also manually clone the options or even have a clone/copy method on it.

uchagani:
@yury-s added back OptionsFactory to define custom options

playwright/src/test/java/com/microsoft/playwright/TestPlaywrightCustomFixtures.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@UsePlaywright(TestPlaywrightCustomFixtures.CustomOptions.class)
+public class TestPlaywrightCustomFixtures {
+  public static class CustomOptions implements OptionsFactory {
+    @Override
+    public Options getOptions() {
+      return new Options().setBaseUrl("https://bing.com").setBrowserName("firefox");
```
yury-s, line 18: Let's use a local url, so that we don't depend on external servers in the tests. Something like `server.EMPTY_PAGE` from other tests should work.

playwright/src/test/java/com/microsoft/playwright/TestPlaywrightCustomOptionFixtures.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@UsePlaywright(TestPlaywrightCustomOptionFixtures.CustomOptions.class)
+public class TestPlaywrightCustomOptionFixtures {
+  public static class CustomOptions implements OptionsFactory {
+    @Override
+    public Options getOptions() {
+      return new Options().setChannel("chrome").setApiRequestOptions(new APIRequest.NewContextOptions().setBaseURL("https://bing.com")).setContextOption(new Browser.NewContextOptions().setBaseURL("https://microsoft.com"));
```
yury-s, line 18: ditto

yury-s (approved):

uchagani:
@yury-s updated tests to use local url

yury-s (approved):

Issue #939, opened 2022-05-27
[Feature] Add emulation config examples to documentation


DonaldTrumpAgent:
Unlike Playwright Nodejs that has a `playwright.config.ts` for presetting all the emulations.
The current documentation of Playwright Java has to use the _JUnit test runners_ in order to configure different user agents.
However, the current example contains only the scenario where the test is only use one browser. 
Would it be possible to add examples with different devices and browsers to the documentation ?

PR #1472, opened 2024-02-03
feat: support device name in playwright fixtures

yury-s:
Reference https://github.com/microsoft/playwright-java/issues/1369
Reference https://github.com/microsoft/playwright-java/issues/939

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -14,7 +14,8 @@ public class Options {
   public ViewportSize viewportSize;
   public String channel;
   public Boolean headless;
-  public String browserName = "chromium";
+  public String browserName;
+  public String deviceName;
   public BrowserType.LaunchOptions launchOptions;
   public Browser.NewContextOptions contextOption;
   public APIRequest.NewContextOptions apiRequestOptions;
@@ -80,6 +81,15 @@ public class Options {
 
   public Options setBrowserName(String browserName) {
     this.browserName = browserName;
+    return this;
+  }
+
+  public String getDeviceName() {
+    return deviceName;
+  }
+
+  public Options setDeviceName(String deviceName) {
+    this.deviceName = deviceName;
     return this;
   }
 
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/DeviceDescriptor.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/DeviceDescriptor.java
@@ -1,0 +1,34 @@
+package com.microsoft.playwright.junit.impl;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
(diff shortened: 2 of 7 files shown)


yury-s:
cc @uchagani

playwright/src/main/java/com/microsoft/playwright/junit/Options.java
```diff
@@ -83,6 +84,15 @@ public Options setBrowserName(String browserName) {
     return this;
   }
 
+  public String getDeviceName() {
```
dgozman, line 87: It's unfortunate that we cannot read back the actual values here, only the device name. However, that's better than nothing.

dgozman (approved):

playwright/src/main/java/com/microsoft/playwright/junit/impl/DeviceDescriptor.java
```diff
@@ -0,0 +1,34 @@
...
+class DeviceDescriptor {
+  public String userAgent;
+  public ViewportSize viewport;
+  public Double deviceScaleFactor;
+  public Boolean isMobile;
+  public Boolean hasTouch;
+  public String defaultBrowserType;
+
+  static DeviceDescriptor findByName(Playwright playwright, String name) {
+    JsonArray devices = ((PlaywrightImpl) playwright).deviceDescriptors();
```
uchagani, line 20: @yury-s  this is returning null for me and so a null pointer exception is being thrown.

uchagani: image

yury-s: Hmm, that's odd. Is `options` in `options.deviceName` null or where does it come from?

uchagani: I believe `PlaywrightImpl.deviceDescriptors()` is returning null so the for loop throws a null pointer exception:

image

yury-s: Hmm, do you have a test where this is happening? This should never happen if playwright is initialized.

yury-s: Ok, I assume it is just the new test. It passes on the bots and locally. Could it be that you have some stale driver? We changed the way device descriptors pushed to the client some time ago. Try downloading [current driver](https://github.com/microsoft/playwright-java/blob/main/CONTRIBUTING.md#updating-driver-version) and see if it still fails.

yury-s: Just running `./scripts/download_driver_for_all_platforms.sh -f` and rebuilding should be sufficient

uchagani: it was failing on main for me too but you figured it out.  downloaded the current driver and it's all good.  apologies for that.  i'll make running that command part of my workflow.

PR #1477, opened 2024-02-07
test: do not rely on channel: chrome

mxschmitt:
Motivation: There is no Google Chrome for `linux-arm64`.

Relates https://github.com/microsoft/playwright-browsers/pull/872.


yury-s (approved):
Let's readd tests that would test the top level options and only run on appropriate bots (e.g. skipped if an env variable is not present).

PR #1483, opened 2024-02-09
devops: do not ship .m2 directory

mxschmitt:
The `.m2` directory contains not only the production dependencies, it also contains the development dependencies in order to develop Playwright.


yury-s (commented):
This contains cached playwright packages and it's dependencies, if the user is going to use the image for running playwright java with the matching version (which is the primary intent of the image), they will now have to redownload all those packages. Not sure we want that.

PR #1487, opened 2024-02-10
remove JUnit Options getters

jfgreffier:
Following https://github.com/microsoft/playwright-java/issues/1369#issuecomment-1924579887

I don't have a strong opinion on letting the fields public or set them private. So this PR can be discussed, or even dropped.


uchagani:
perfect thank you for doing that.  I agree with removing them to keep the same standard as the other options in playwright-java

yury-s (approved):

PR #1491, opened 2024-02-12
feat(junit): testIdAttribute option

yury-s:
Reference https://github.com/microsoft/playwright-java/issues/1369


playwright/src/main/java/com/microsoft/playwright/junit/impl/BrowserExtension.java
```diff
@@ -42,6 +42,9 @@ static Browser getOrCreateBrowser(ExtensionContext extensionContext) {
     Playwright playwright = PlaywrightExtension.getOrCreatePlaywright(extensionContext);
     BrowserType.LaunchOptions launchOptions = getLaunchOptions(options);
 
+    String testIdAttribute = options.testIdAttribute == null ? "data-testid" : options.testIdAttribute;
+    playwright.selectors().setTestIdAttribute(testIdAttribute);
```
dgozman, line 46: Since it is easy for us to change this option between tests, should we move this code to the context creation place? Additionally, we can copy it to `getOrCreatePlaywright()` to accommodate custom contexts/browsers created by the user.

PR #1492, opened 2024-02-14
feat(junit): Close Playwright after test run

uchagani:
This PR adds the ability to close Playwright objects after the entire test run.  I also added a `ParameterResolver` for server so we don't have to create methods to access it from the `serverMap`.

Ref: https://github.com/microsoft/playwright-java/issues/1369#issuecomment-1924579887

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
@@ -5,6 +5,7 @@ import com.microsoft.playwright.junit.Options;
 import org.junit.jupiter.api.extension.*;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -22,7 +23,7 @@ public class PlaywrightExtension implements ParameterResolver, BeforeAllCallback
     isTestRunStarted = false;
     beforeLock = new ReentrantLock();
     threadLocalPlaywright = new ThreadLocal<>();
-    playwrightList = new ArrayList<>();
+    playwrightList = Collections.synchronizedList(new ArrayList<>());
   }
 
   // Before a Test class starts, we register a closeable resource
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
@@ -18,6 +18,7 @@ public class PlaywrightExtension implements ParameterResolver, BeforeAllCallback
   private static final List<Playwright> playwrightList;
   private final static Lock beforeLock;
   private static boolean isTestRunStarted;
+  private static final ExtensionContext.Namespace namespace = ExtensionContext.Namespace.create(PlaywrightExtension.class);
 
   static {
     isTestRunStarted = false;
@@ -34,7 +35,7 @@ public class PlaywrightExtension implements ParameterResolver, BeforeAllCallback
       beforeLock.lock();
       if (!isTestRunStarted) {
         isTestRunStarted = true;
-        context.getRoot().getStore(ExtensionContext.Namespace.GLOBAL).put(context.getUniqueId(), this);
+        context.getRoot().getStore(namespace).put(context.getUniqueId(), this);
       }
     } finally {
       beforeLock.unlock();
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
@@ -46,10 +46,9 @@ public class PlaywrightExtension implements ParameterResolver, BeforeAllCallback
   @Override
   public void close() throws Throwable {
     for (Playwright playwright : playwrightList) {
-      if (playwright != null) {
-        playwright.close();
-      }
+      playwright.close();
     }
+    playwrightList.clear();
   }
 
   @Override
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
@@ -4,18 +4,50 @@ import com.microsoft.playwright.Playwright;
 import com.microsoft.playwright.junit.Options;
 import org.junit.jupiter.api.extension.*;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
 import static com.microsoft.playwright.junit.impl.ExtensionUtils.isParameterSupported;
 import static com.microsoft.playwright.junit.impl.ExtensionUtils.setTestIdAttribute;
 
-public class PlaywrightExtension implements ParameterResolver, AfterAllCallback {
-  private static final ThreadLocal<Playwright> threadLocalPlaywright = new ThreadLocal<>();
+public class PlaywrightExtension implements ParameterResolver, BeforeAllCallback, ExtensionContext.Store.CloseableResource {
+  private static final ThreadLocal<Playwright> threadLocalPlaywright;
+  private static final List<Playwright> playwrightList;
+  private final static Lock beforeLock;
+  private static boolean isTestRunStarted;
 
+  static {
+    isTestRunStarted = false;
+    beforeLock = new ReentrantLock();
+    threadLocalPlaywright = new ThreadLocal<>();
+    playwrightList = new ArrayList<>();
+  }
+
+  // Before a Test class starts, we register a closeable resource
+  // We use a lock + boolean to ensure this only gets run once for the entire test run
   @Override
-  public void afterAll(ExtensionContext extensionContext) {
-    Playwright playwright = threadLocalPlaywright.get();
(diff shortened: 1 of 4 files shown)


playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
```diff
@@ -4,18 +4,50 @@
...
+  private static final ThreadLocal<Playwright> threadLocalPlaywright;
+  private static final List<Playwright> playwrightList;
+  private final static Lock beforeLock;
+  private static boolean isTestRunStarted;
+
+  static {
+    isTestRunStarted = false;
+    beforeLock = new ReentrantLock();
+    threadLocalPlaywright = new ThreadLocal<>();
+    playwrightList = new ArrayList<>();
```
yury-s, line 25: It should be `Collections.synchronizedList` as there is a chance it is accessed from multiple threads concurrently.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
```diff
@@ -4,18 +4,50 @@
...
+    } finally {
+      beforeLock.unlock();
+    }
+  }
+
+  // This is a workaround for JUnit's lack of an "AfterTestRun" hook
+  // This will be called once after all tests have completed.
+  @Override
+  public void close() throws Throwable {
+    for (Playwright playwright : playwrightList) {
```
yury-s, line 47: This way we close playwright instance that is still referenced from a thread local. To avoid that we could have a pool of playwright instances, so that whenever a test on a thread starts, it borrows a spare playwright from the pool or creates a new one. After all test methods have finished it must return the playwright instance to the pool and clear thread local reference to it. If next test starts on the same thread, it will borrow playwright from the pool again. This assumes that all tests within the class run on the same thread (which I believe is our assumption in other places too. After all tests have completed there should be no thread local references to the playwright instances and we can safely  close all of them. What do you think?

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
uchagani: makes sense i will update it.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
uchagani: > This way we close playwright instance that is still referenced from a thread local. 

my understanding is that ThreadLocals are closed when the thread is killed so do we need to worry about this?  This close method will only run after all tests have run so all threads are done at that point and no future threads will launch.  Is your concern that closing playwright here will fail?  I tested this and didn't see any exceptions being thrown.  Apologies for being dense here but i don't understand the concern.

I'm not opposed to doing some sort pool for it but I think it will take more work.  Maybe we can do it in a future iteration?

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
uchagani: done

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
yury-s: > my understanding is that ThreadLocals are closed when the thread is killed so do we need to worry about this?

Yes, you are right. Java docs state that [explicitly](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) " after a thread goes away, all of its copies of thread-local instances are subject to garbage collection (unless other references to these copies exist)." Nothing to worry about then, we can collect them in a list and close after all tests are done. There should be no usages after that point, unless the tests posted a reference that will be accessed on a different thread, so it should be safe to close for us.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
```diff
@@ -4,18 +4,51 @@
...
+      if (playwright != null) {
```
yury-s, line 49: `playwright` is never null here, we can drop the check?

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
```diff
@@ -4,18 +4,51 @@
...
+        playwright.close();
+      }
```
yury-s, line 51: let's reset/clear the list after closing, to avoid any hanging references

yury-s (approved):

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
```diff
@@ -4,19 +4,51 @@
...
+  // Before a Test class starts, we register a closeable resource
+  // We use a lock + boolean to ensure this only gets run once for the entire test run
   @Override
-  public void afterAll(ExtensionContext extensionContext) {
-    Playwright playwright = threadLocalPlaywright.get();
-    threadLocalPlaywright.remove();
-    if (playwright != null) {
+  public void beforeAll(final ExtensionContext context) {
+    try {
+      beforeLock.lock();
```
yury-s, line 34: Correct me if I am wrong, but as far as I understand we don't need neither beforeLock nor isTestRunStarted here because PlaywrightExtension instance is created once per test class and for that test class `beforeAll` is also called exactly once. `ExtensionContext.Store` should already handle concurrent access as it can be access by multiple tests. Their [source code](https://github.com/sbrannen/junit5/blob/84ac590dd9bdee656bfbfd45aa7afbdfd26a3478/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java#L49
) matches my understanding.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
```diff
@@ -4,19 +4,51 @@
...
+      if (!isTestRunStarted) {
+        isTestRunStarted = true;
+        context.getRoot().getStore(ExtensionContext.Namespace.GLOBAL).put(context.getUniqueId(), this);
```
yury-s, line 37: Let's use Playwright specific namespace here to avoid possible clashes? Something like `ExtensionContext.Namespace.create(PlaywrightException.class)` should do.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
uchagani: You are correct that beforeAll is called once per test class, however we want this particular beforeAll to be called once per test run and register the closeable resource once.

The lock is to ensure if two test classes hit the beforeAll at the same time, only one of them can register the closeable resource.

If we don't do it this way, every test class will register its own closeable resource and the close method will be called after each test class completes, which is essentially an AfterAll, and that's what we don't want to do.

We want to close all playwright instances after the entire test run is complete.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
uchagani: I'll need to look into this some more but I believe that closeable resources need to be registered in the global namespace. I don't think there will be conflicts with anything else because we are using this extension's context's unique id as the key.

But I'll play around with it to see if we can register a closeable resource in a named namespace.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
yury-s: I see what you mean. In that case how about we introduce a class that manages the lifetime of the Playwright objects and make it closable:

```java
  class PlaywrgihtInstances implements ExtensionContext.Store.CloseableResource {
    private List<Playwright> playwrightList;

    static synchronized PlaywrgihtInstances getOrCreateFor(ExtensionContext.Store rootStore) {
      if (rootStore.get(PlaywrgihtInstances.class) != null) {
        rootStore.put(PlaywrgihtInstances.class, new PlaywrgihtInstances());
      }
    }

    @Override
    public void close() throws Throwable {
      ...
    }
  }
```

The you just need to create playwrights via this class to make sure all of them are registered right after creation. WDYT?

uchagani: i feel like this would make it less performant.  getOrCreate can be called many different times over a test class.  if a test class 5 methods, it could be called 5 times.  where as the beforeAll is only called once per test class and never more.

maybe i'm not seeing the benefit of this.  would you mind implementing what you are suggesting in this branch.  following the code would make it easier for me to understand.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
uchagani: @yury-s verified that closeable resource can be registered in a named namespace so added that as you asked.

playwright/src/main/java/com/microsoft/playwright/junit/impl/PlaywrightExtension.java
yury-s: It'd only need to be called where `playwrightList.add(` is currently called. Basically the same number of times as the number of playwright instances created. I can send a follow-up PR with the change, no problem.

PR #1494, opened 2024-02-15
feat: roll 1.42.0-alpha driver, implement page.addLocatorHandler

yury-s:
Reference https://github.com/microsoft/playwright-java/issues/1456


dgozman (approved):

playwright/src/main/java/com/microsoft/playwright/impl/PageImpl.java
```diff
@@ -523,6 +528,34 @@ public List<ElementHandle> querySelectorAll(String selector) {
...
+  private void onLocatorHandlerTriggered(int uid) {
+    try {
+      Runnable handler = locatorHandlers.get(uid);
+      if (handler != null) {
+        handler.run();
+      }
+    } finally {
+      JsonObject params = new JsonObject();
+      params.addProperty("uid", uid);
+      sendMessageAsync("resolveLocatorHandlerNoReply", params);
```
mxschmitt, line 555: whats the difference between sendMessage and sendMessageAsync?

mxschmitt (approved):

playwright/src/main/java/com/microsoft/playwright/impl/PageImpl.java
yury-s: The latter is not waiting for the response. One consequence is that it also doesn't throw.

PR #1506, opened 2024-03-08
Convert TestBrowser to use fixtures

uchagani:
Converting existing tests to use fixtures.
Ref: #1369


playwright/src/test/java/com/microsoft/playwright/junit/TestBrowser2.java
```diff
@@ -0,0 +1,20 @@
...
+import static com.microsoft.playwright.junit.TestOptionsFactories.ChannelOptionsFactory.getBrowserChannelEnumFromEnv;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@UsePlaywright(TestOptionsFactories.ChannelOptionsFactory.class)
+public class TestBrowser2 {
+
+  @Test
+  void shouldSupportDeprecatedChannelEnum(Browser browser) {
+    BrowserChannel channel = getBrowserChannelEnumFromEnv();
+    Assumptions.assumeTrue(channel != null);
```
yury-s, line 17: This is too, the `browser` will have been created by the time we check this assumption here. Can we move this call into ChannelOptionsFactory.getOptions and see if the test is skipped if there is no channel? I think it's fine to write this test without the `Browser` fixture (just call the utility methods directly), it would probably make it's purpose more clear. You could still use `Playwright` fixture though.

playwright/src/test/java/com/microsoft/playwright/junit/TestOptionsFactories.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public static class BasicOptionsFactory implements OptionsFactory {
+    @Override
+    public Options getOptions() {
+      return new Options().setBrowserName(getBrowserName());
+    }
+  }
+
+  public static class ChannelOptionsFactory implements OptionsFactory {
```
yury-s, line 42: Let's move this factory definition into the only class that depends on it?

playwright/src/test/java/com/microsoft/playwright/junit/TestBrowser1.java
yury-s, line 1: I guess the idea is to place tests that use fixtures into a separate package. In that case I'd keep them in a different package than the tests that check fixtures themselves, keeping them in the same playwright/src/test/java/com/microsoft/playwright/ is also fine by me.

yury-s (approved):

playwright/src/test/java/com/microsoft/playwright/junit/TestBrowser1.java
uchagani: I was back and forth on this.  I was going to leave the old TestBrowser in the project but then decided to delete it.  Now that its deleted (and if you're okay with it), I think moving the new tests into `playwright/src/test/java/com/microsoft/playwright/ ` makes sense to me

uchagani:
@yury-s, as i was writing these tests a thought came to me:  Because we use 1 browser per thread, and currently only support concurrency per class, if a user closes the browser in a test, and then uses the browser fixture in another test, they will get a closed browser.  Should we have something in the `BrowserExtension` to detect/handle this?

```java
@UsePlaywright
public class FooTests {

  @Test
  void foo(Browser browser) {
    browser.close();
  }

  @Test
  void foo1(Browser browser) {
    //browser will be closed here
  }

}
```

yury-s:
> @yury-s, as i was writing these tests a thought came to me: Because we use 1 browser per thread, and currently only support concurrency per class, if a user closes the browser in a test, and then uses the browser fixture in another test, they will get a closed browser. Should we have something in the `BrowserExtension` to detect/handle this?
> 
> ```java
> @UsePlaywright
> public class FooTests {
> 
>   @Test
>   void foo(Browser browser) {
>     browser.close();
>   }
> 
>   @Test
>   void foo1(Browser browser) {
>     //browser will be closed here
>   }
> 
> }
> ```

Good catch. I think we could add `Browser. onDisconnected` handler and delete the browser from the fixture's cache if it closes. That way next test should automatically get a valid browser.

PR #1515, opened 2024-03-14
fix(junit): Fixes fixtures not working in nested junit test classes

### LINKED ISSUE #1514: [Feature]: Add @Inherited to @UsePlaywright
### 🚀 Feature Request

Hi there,

Im playing around with the new Junit integration and while it works as expected we cannot use it right now within our tests.
We use some abstract super classes and also `@Nested` classes to group scenarios. `@UsePlaywright` currently is not `@Inherited` and therefore during extension disovery it fails at this place `com.microsoft.playwright.junit.impl.ExtensionUtils#hasUsePlaywrightAnnotation` 

### Example

_No response_

### Motivation

```
@UsePlaywright
public class TestExample {
  @Nested
  class Scenario {  
    @Test
    void shouldCheckTheBox(Page page) {
      page.setContent("<input id='checkbox' type='checkbox'></input>");
      page.locator("input").check();
      assertEquals(true, page.evaluate("window['checkbox'].checked"));
    }
  }
}
```

uchagani:
Resolves #1514 

Allows fixtures to be used inside nested class and allows inheriting fixtures from base classes.

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
@@ -54,7 +54,7 @@ class ExtensionUtils {
   }
 
   private static Class<?> getOuterClass(Class<?> clazz) {
-    if (clazz.getEnclosingClass() == null) {
+    if (clazz.getDeclaringClass() == null) {
       return clazz;
     } else {
       return getOuterClass(clazz.getDeclaringClass());
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
@@ -54,7 +54,7 @@ class ExtensionUtils {
   }
 
   private static Class<?> getOuterClass(Class<?> clazz) {
-    if (clazz.getDeclaringClass() == null) {
+    if (clazz.getEnclosingClass() == null) {
       return clazz;
     } else {
       return getOuterClass(clazz.getDeclaringClass());
--- a/playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
@@ -27,11 +27,13 @@ import static org.junit.platform.commons.support.AnnotationSupport.findAnnotatio
 
 class ExtensionUtils {
   static boolean hasUsePlaywrightAnnotation(ExtensionContext extensionContext) {
-    return AnnotationSupport.isAnnotated(extensionContext.getTestClass(), UsePlaywright.class);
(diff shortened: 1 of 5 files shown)


playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
```diff
@@ -49,6 +51,13 @@ static boolean isParameterSupported(ParameterContext parameterContext, Extension
   static void setTestIdAttribute(Playwright playwright, Options options) {
     String testIdAttribute = options.testIdAttribute == null ? "data-testid" : options.testIdAttribute;
     playwright.selectors().setTestIdAttribute(testIdAttribute);
+  }
 
+  private static Class<?> getOuterClass(Class<?> clazz) {
+    if (clazz.getDeclaringClass() == null) {
```
yury-s, line 57: Should it be `getEnclosingClass` to work for anonymous nested classes too?

yury-s (approved):

playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
uchagani: I don't think you can have anonymous inner classes as test classes. I could be wrong though.

uchagani: I can update it, I don't think it hurts anything

yury-s: > I don't think you can have anonymous inner classes as test classes. I could be wrong though.

Now that I think about it, I believe you are right. The only way to declare an anonymous class would be to create an instance of it, but one won't be able to mark it as @Nested.

playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
```diff
@@ -49,6 +51,13 @@ static boolean isParameterSupported(ParameterContext parameterContext, Extension
   static void setTestIdAttribute(Playwright playwright, Options options) {
     String testIdAttribute = options.testIdAttribute == null ? "data-testid" : options.testIdAttribute;
     playwright.selectors().setTestIdAttribute(testIdAttribute);
+  }
 
+  private static Class<?> getOuterClass(Class<?> clazz) {
+    if (clazz.getEnclosingClass() == null) {
+      return clazz;
+    } else {
+      return getOuterClass(clazz.getDeclaringClass());
```
yury-s, line 60: Let's use either getDeclaringClass() or getEnclosingClass() in both branches, otherwise it might throw NPE.

yury-s (approved):

playwright/src/main/java/com/microsoft/playwright/junit/impl/ExtensionUtils.java
uchagani: ah i missed that, thanks for catching/fixing

PR #1521, opened 2024-03-21
chore(test): Convert tests to use fixtures

uchagani:
@yury-s, converted more tests to use fixtures

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
@@ -16,32 +16,37 @@
 
 package com.microsoft.playwright;
 
-import org.junit.jupiter.api.Disabled;
+import com.microsoft.playwright.junit.FixtureTest;
+import com.microsoft.playwright.junit.UsePlaywright;
 import org.junit.jupiter.api.Test;
 
 import java.io.OutputStreamWriter;
 import java.util.ArrayList;
 import java.util.List;
 
+import static com.microsoft.playwright.TestOptionsFactories.isFirefox;
+import static com.microsoft.playwright.TestOptionsFactories.isWebKit;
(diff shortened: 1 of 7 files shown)


playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
```diff
@@ -16,32 +16,37 @@
...
+  void shouldCreateNewContext(Browser browser) {
+    assertEquals(0, browser.contexts().size());
     BrowserContext context = browser.newContext();
-    assertEquals(2, browser.contexts().size());
+    assertEquals(1, browser.contexts().size());
     assertTrue(browser.contexts().indexOf(context) != -1);
     assertEquals(context.browser(), browser);
     context.close();
-    assertEquals(1, browser.contexts().size());
+    assertEquals(0, browser.contexts().size());
```
uchagani, line 44: these assertions needed to be updated because the TestBase used to create a context but fixtures don't create one unless you ask for it.

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
```diff
@@ -92,11 +97,11 @@ void shouldIsolateLocalStorageAndCookies() {
     context1.close();
     context2.close();
 
-    assertEquals(1, browser.contexts().size());
+    assertEquals(0, browser.contexts().size());
```
uchagani, line 100: these assertions needed to be updated because the TestBase used to create a context but fixtures don't create one unless you ask for it.

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextBasic.java
yury-s: We need to be careful to close contexts if (another) test fails, otherwise this might fail as a side effect.

yury-s (approved):

PR #1541, opened 2024-04-06
feat(junit): add ability to connect to remote browsers via fixtures

### LINKED ISSUE #1537: [Feature]: Allow user to pass in wsEndpoint and ConnectOptions to fixtures
### 🚀 Feature Request

I would like to be able to specify a `wsEndpoint` and `ConnectOptions` in the `OptionsFactory` and use the `connect` method to get a `Browser`.  Similar to https://playwright.dev/docs/api/class-testoptions#test-options-connect-options

### Example

options can be defined like this:

```java
class PlaywrightOptions implements OptionsFactory {
    @Override
    public Options getOptions() {
      return new Options()
        .setWsEndpoint("ws://localhost:5678")
        .setConnectOptions(new BrowserType.ConnectOptions().setTimeout(10000));
    }
}
```

The `BrowserExtension` could check to see if the `wsEndpoint` field is not null and if so, use the connect method to create a `Browser` object instead of the `launch` method.

### Motivation

The fixtures need to be able to use this option in order to run against cloud services like BrowserStack or Saucelabs.

uchagani:
Resolves #1537

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/junit/OptionsExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/junit/OptionsExtension.java
@@ -25,7 +25,8 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 import static com.microsoft.playwright.impl.junit.ExtensionUtils.*;
 
 public class OptionsExtension implements AfterAllCallback {
-  private static final ThreadLocal<Options> threadLocalOptions = new ThreadLocal<>();
+  // This is public to allow our tests to set the wsEndpoint dynamically
+  public static final ThreadLocal<Options> threadLocalOptions = new ThreadLocal<>();
 
   @Override
   public void afterAll(ExtensionContext extensionContext) {
--- a/playwright/src/main/java/com/microsoft/playwright/impl/junit/OptionsExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/junit/OptionsExtension.java
@@ -25,8 +25,7 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 import static com.microsoft.playwright.impl.junit.ExtensionUtils.*;
 
 public class OptionsExtension implements AfterAllCallback {
-  // This is public to allow our tests to set the wsEndpoint dynamically
-  public static final ThreadLocal<Options> threadLocalOptions = new ThreadLocal<>();
+  private static final ThreadLocal<Options> threadLocalOptions = new ThreadLocal<>();
 
   @Override
   public void afterAll(ExtensionContext extensionContext) {
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -43,6 +43,7 @@ public class Options {
   public Browser.NewContextOptions contextOptions;
   public APIRequest.NewContextOptions apiRequestOptions;
   public Playwright.CreateOptions playwrightCreateOptions;
+  // WebSocket endpoint to be used when connecting to a remote browser.
   public String wsEndpoint;
   public BrowserType.ConnectOptions connectOptions;
 
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -43,6 +43,18 @@ public class Options {
   public Browser.NewContextOptions contextOptions;
   public APIRequest.NewContextOptions apiRequestOptions;
   public Playwright.CreateOptions playwrightCreateOptions;
+  public String wsEndpoint;
+  public BrowserType.ConnectOptions connectOptions;
+
+  public Options setWsEndpoint(String wsEndpoint) {
+    this.wsEndpoint = wsEndpoint;
+    return this;
+  }
+
+  public Options setConnectOptions(BrowserType.ConnectOptions connectOptions) {
+    this.connectOptions = connectOptions;
+    return this;
+  }
 
   public Options setPlaywrightCreateOptions(Playwright.CreateOptions playwrightCreateOptions) {
     this.playwrightCreateOptions = playwrightCreateOptions;
--- a/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
+++ b/playwright/src/main/java/com/microsoft/playwright/junit/Options.java
@@ -44,6 +44,7 @@ public class Options {
   public APIRequest.NewContextOptions apiRequestOptions;
   public Playwright.CreateOptions playwrightCreateOptions;
   // WebSocket endpoint to be used when connecting to a remote browser.
+  // If this is set, BrowserType.connect will be used.  Otherwise, BrowserType.launch will be used.
   public String wsEndpoint;
   public BrowserType.ConnectOptions connectOptions;
 
--- a/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
@@ -16,6 +16,7 @@
 
 package com.microsoft.playwright;
 
+import com.microsoft.playwright.impl.ImplUtils;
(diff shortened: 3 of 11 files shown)


uchagani:
@yury-s, I wanted to add a test to ensure that the `Browser` returned from the fixture is actually a browser that connects via web socket but I couldn't find a clean way to do that.  `Browser#isConnected` returns true regardless of if the browser used `launch` or `connected`.  There is a `isConnectedOverWebSocket` but that is in `BrowserImpl` and I wouldn't have access to that in the test.  Any thoughts on how we can test this?

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
```diff
@@ -34,16 +36,19 @@
...
 
-public class TestBrowserTypeConnect extends TestBase {
+@FixtureTest
+@UsePlaywright(TestBrowserTypeConnect.ConnectOptionsFactory.class)
+public class TestBrowserTypeConnect {
   static Path FILE_TO_UPLOAD = Paths.get("src/test/resources/file-to-upload.txt");
 
-  private Process browserServer;
-  private String wsEndpoint;
+  private static Process browserServer;
```
yury-s, line 50: What's the rationale for changing these to public? These tests are independent and (at least in theory) could run in parallel but if we the browserServer/endpoint are shared in a static field, it may lead to races (admittedly, not in the current mode where we don't parallelize within class).

playwright/src/main/java/com/microsoft/playwright/impl/junit/OptionsExtension.java
```diff
@@ -25,7 +25,8 @@
 import static com.microsoft.playwright.impl.junit.ExtensionUtils.*;
 
 public class OptionsExtension implements AfterAllCallback {
-  private static final ThreadLocal<Options> threadLocalOptions = new ThreadLocal<>();
+  // This is public to allow our tests to set the wsEndpoint dynamically
```
yury-s, line 28: Not sure, I understand why this needs to be public.

yury-s:
>  There is a `isConnectedOverWebSocket` but that is in `BrowserImpl` and I wouldn't have access to that in the test. Any thoughts on how we can test this?

How about creating an utils class under `playwright/src/test` that would expose the field to the tests:

```java
package com.microsoft.playwright.impl;

import com.microsoft.playwright.Browser;

public class ImplUtils {
  public static boolean isRemoteBrowser(Browser browser) {
    return ((BrowserImpl) browser).isConnectedOverWebSocket;
  }
}
```

would that work?

playwright/src/main/java/com/microsoft/playwright/junit/Options.java
```diff
@@ -43,6 +43,18 @@ public class Options {
   public Browser.NewContextOptions contextOptions;
   public APIRequest.NewContextOptions apiRequestOptions;
   public Playwright.CreateOptions playwrightCreateOptions;
+  public String wsEndpoint;
```
yury-s, line 46: Add a comment?

playwright/src/main/java/com/microsoft/playwright/impl/junit/OptionsExtension.java
uchagani: The way the BrowserConnect tests are written, the wsEndpoint isn't known until after the Options are already created by the fixture.  Making this public is the only way i could think of to set the wsEndpoint in the BeforeAll of the test class.  I guess I could use reflection to do it.

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
uchagani: do you mean why did I make them static?  BeforeAll/AfterAll are supposed to be static by default and because these are set/closed in those methods, I made these static also.  https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-instance-lifecycle

uchagani: I can add `@TestInstance(Lifecycle.PER_CLASS)` to this test class and change the fields back to being instance fields if you'd like.  I was trying to keep as much default junit functionality as possible.  As far as supporting per-method parallelization goes, we would either keep these as static or we would be launching/closing the browser in `BeforeEach/AfterEach` hooks.

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
```diff
@@ -85,22 +91,24 @@ private static BrowserServer launchBrowserServer(BrowserType browserType) {
...
-  // Hide base class method to launch browser server and connect to it.
-  void launchBrowser() {
-    initBrowserType();
+  static void launchBrowser(BrowserType browserType) {
     BrowserServer r = launchBrowserServer(browserType);
     wsEndpoint = r.wsEndpoint;
     browserServer = r.process;
-    browser = browserType.connect(wsEndpoint);
-    // Do not actually connect to browser, the tests will do it manually.
+    OptionsExtension.threadLocalOptions.set(new ConnectOptionsFactory().getOptions().setWsEndpoint(wsEndpoint));
```
uchagani, line 107: @yury-s this is why the `ThreadLocal<Options>` needed to be public.  The `wsEndpoint` isn't known until the `BeforeAll` hook is executed, at that point the `Options` object has already been created.  We needed a way to dynamically alter this.  

I am not advocating for this behavior, it's just what I saw as necessary.  We can make it private again and I can set this option value via reflection.

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
yury-s: It was inheriting  `@TestInstance(Lifecycle.PER_CLASS)` from `TestBase` before, so I'd just add the annotation explicitly while switching to `@UsePlaywright`.

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
yury-s: I don't like us writing into variables which are very specific to the implementation of the fixtures. Can we instead lazily launch the browser server from inside `ConnectOptionsFactory.getOptions` when creating the `Options` instance?

Alternatively, we can leave existing `TestBrowserTypeConnect` test alone and create a new one under `./tools`. We could launch the browser server separately and then run a playwright test that populates `ws` address in UsePlaywright options with e.g. an environment variable value. We are testing a very specific setup with the remote connection and I believe having a dedicated test would be justified in this case. WDYT?

uchagani: > Alternatively, we can leave existing TestBrowserTypeConnect test alone and create a new one under ./tools. We could launch the browser server separately and then run a playwright test that populates ws address in UsePlaywright options with e.g. an environment variable value. We are testing a very specific setup with the remote connection and I believe having a dedicated test would be justified in this case. WDYT?

I like this but there is a lot going on in the `launchBrowser` methods that I don't totally understand.  I would need some help with how that would work.

But this actually gives me another idea.  We can create another internal extension to launch the browser and set a variable with the wsEndpoint.  I'll push up what I'm thinking for your review.

yury-s: Sure, I can help with that. Let's revert changes in this file and in OptionsExtension.java and land the rest without a test for now. We can add it later while letting you use it in the other project. We can also not add a test for now if it implies a lot of extra complexity. I don't think we have such test for Node.js, but we run it with the service, could do the same here.

yury-s (approved):

playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
uchagani: Sounds good to me.  I've reverted the tests and extension changes.

yury-s (approved):

Issue #1548, opened 2024-04-15
[Bug]: PLAYWRIGHT_NODEJS_PATH env variable is ignored in version 1.43.0


marrek13:
### Version

1.43.0

### Steps to reproduce

Use custom PLAYWRIGHT_NODEJS_PATH env and try to run the application, it is now ignored and playwright java tries to use a node executable from a non-existing /tmp directory

### Expected behavior

PLAYWRIGHT_NODEJS_PATH is respected as previously

### Actual behavior

PLAYWRIGHT_NODEJS_PATH is ignored

### Additional context

Exception:

```json
{
   "@timestamp":"2024-04-15T08:17:46.483296314Z",
   "@version":"1",
   "message":"Application run failed",
   "logger_name":"org.springframework.boot.SpringApplication",
   "thread_name":"main",
   "level":"ERROR",
   "level_value":40000,
[...]

PR #1552, opened 2024-04-16
fix(driver): consider PLAYWRIGHT_NODEJS_PATH from host env

mxschmitt:
Fixes https://github.com/microsoft/playwright-java/issues/1548


yury-s (approved):
Do you mind adding a test? We already have one that checks setting this variable in Playwright.create and it did pass after the change.

marrek13:
Is there a chance to release a bugfix release with this fix?

PR #1560, opened 2024-04-29
feat(junit): Implement automatic saving of traces and screenshots via fixtures

### LINKED ISSUE #1526: [Feature]: Allow automatic saving of trace files
### 🚀 Feature Request

This is also kind of a bug but I decided to make it a feature request.  

Feature Request: I would like the ability to automatically save trace files.  

Bug:  I cannot save trace file based on test status.

### Example

An option in `OptionsFactory` that allows the user to configure when to save trace files `off`, `on`, `retain-on-failure` and an option for a root trace directory.  The fixtures would then automatically start the trace, and depending on the option and the test result, would save the trace files in the root trace directory.  

### Motivation

As I said, this is a feature request/bug report.  I think the benefits are clear but the bug needs to be explained.

Currently, the user can only save a trace file by doing something like this:

```java
@AfterEach
void afterEach(BrowserContext browserContext) {
  // save trace file
}
```

There is no way for the user to only save trace files if the test fails (`retain-on-failure`).  JUnit recommends implementing the [TestWatcher interface](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/TestWatcher.html).  The problem is that the test watcher methods are called after the `AfterEachCallback` methods, so when the test watcher method is called, the `BrowserContext` is already closed by the `BrowserContextExtension`.  So even if I use reflection to get at the `BrowserContext` inside my `TestWatcher`, an exception is thrown.

If we allow the fixtures to handle traces this issue can be easily solved

uchagani:
Resolves #1526.  This PR also adds saving screenshots.

I tried to keep the API of the fixture-generated artifacts as close to node as possible.  

Traces:
By default, if no `outputDir` is specified in the `Options`, the default output path will be `projectDir/test-results/fully.qualified.ClassName.testMethodName-browserName/trace.zip`.  If an `outputDir` is specified in the `Options` then the artifacts will go in `userSpecifiedOutputDir/fully.qualified.ClassName.testMethodName-browserName/trace.zip`.

Screenshots:
Screenshots follow the same nomenclature as the traces with the exception that screenshot files are named `test-finished-1.png`.  Depending on the number of pages a `BrowserContext` has, the `1` will be incremented per `Page`.

~~I added tests to test this feature.  Since traces and screenshots are created after the test is finished, I needed to add a new test dependency (`junit-platform-testkit`).   This is JUnit's preferred way to test extensions.~~

~~`@Transient` annotation was created because the new tests should not be run by themselves and only run through the `EngineTestKit`.  I updated the GitHub actions to exclude these tests.~~

~~There is some weird threading issues happening when the new tests run with the existing suite so I added a new step in the CI actions to run the new tests separately.~~

problems with running tests through the `junit-platform-testkit`: https://github.com/microsoft/playwright-java/pull/1560#issuecomment-2305981500


uchagani:
I don't think the failures here are related to the PR.  could we try re-running?

yury-s:
> I don't think the failures here are related to the PR. could we try re-running?

Restarted. But it looks like at least some of them are in the new tests.

playwright/src/test/java/com/microsoft/playwright/junit/TestTracesAndScreenshots.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  @Test
+  void testTraceRetainOnFailure() {
+    Class<?> testClass = TestTraceRetainOnFailures.class;
+    List<Method> testMethods = getTestMethods(testClass);
+
+    for (Method test : testMethods) {
+      File traceFile = getTracePathForDefaultLocation(testClass, test).toFile();
+      if (test.getName().contains("ShouldNot")) {
```
yury-s, line 90: I'd check entire method name "traceShouldNotExistWhenTestPasses"  for clarity.

Also here and in other tests, when iterating over several methods, include the name in the message for easier debugging: assertFalse(traceFile.exists(), test.getName());

yury-s (approved):
Looks good, but let's figure out the multithreading issues and other failures of the new tests on the bots.

uchagani:
@yury-s the threading issues, I believe, are resulting from tests running via the engine kit and the normal way at the same time.

I believe the threading issues are due to use using threadlocal in our extensions. Somewhere, playwright is getting closed by the ExtensionContext but there are still tests running on that thread.

We should eventually move away from using threadlocal and store objects in the ExtensionContext (this is what Junit recommends). However, I'm not sure how we can scope to threads while doing that. We can scope to method or class easily but we want to reuse playwright and keep it open across test classes.


I don't think users will be affected by this but will verify by running the whole suite except engine kit tests.

playwright/src/test/java/com/microsoft/playwright/junit/TestTracesAndScreenshots.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static com.microsoft.playwright.TestOptionsFactories.getBrowserName;
+import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+
+// Test both traces and screenshots in the same class as they both have the same default directory
+// and cleanup is easier
+@Tag("fixtureTracing") // These tests should be run in isolation as there are threading issues when run with others
```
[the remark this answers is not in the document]
uchagani: mentioned in the main comments, I think they have to do with tests being run in the EngineTestKit on the same thread as the rest of the tests.  I don't think our users will be affected by this but I will double check.

playwright/src/test/java/com/microsoft/playwright/junit/TestTracesAndScreenshots.java
```diff
@@ -0,0 +1,201 @@
...
+  private List<Method> getTestMethods(Class<?> testClass) {
+    return Arrays.stream(testClass.getDeclaredMethods()).filter(m -> m.isAnnotationPresent(Test.class)).collect(Collectors.toList());
+  }
+
+  private static Path getDefaultOutputPath() {
+    return Paths.get(System.getProperty("user.dir")).resolve("test-results");
+  }
+
+  public static Path getCustomOutputPath() {
+    return Paths.get(System.getProperty("java.io.tmpdir")).resolve("playwright-custom-output");
```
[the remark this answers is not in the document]
uchagani: I tried using this but we need this resolved in the options which is before any test would resolve this, so i chose to do it this way and clean up manually.

uchagani:
@yury-s i believe i have addressed all the comments.  I also verified that the threading issues are not user-facing and should not affect users.  After this PR, I can start looking into how we can store playwright fixtures in the ExtensionContext and scope them to specific threads

uchagani:
will work on test failures this evening

uchagani:
I'm sorry for the delay life got in the way recently.  Hopefully will get to this this weekend

yury-s:
No problem at all, hope everything is alright.

uchagani:
@yury-s was wondering if it's okay to land this feature without tests.  The JUnit Engine Test Kit isn't playing well.  I have confirmed that the issue is not in the feature but how the engine test kit runs the tests.

playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
```diff
@@ -66,10 +64,135 @@ public static BrowserContext getOrCreateBrowserContext(ExtensionContext extensio
...
+    }
+
+    if (options.screenshot.equals(Options.Screenshot.ON) || options.screenshot.equals(Options.Screenshot.ONLY_ON_FAILURE)) {
+      saveScreenshot(extensionContext, true);
+    }
+    cleanupResources();
+  }
+
+  private void cleanupResources() {
+    closePage();
```
yury-s, line 106: Can we rely on `BrowserContext.close()` closing all pages and avoid this call?

playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
```diff
@@ -66,10 +64,135 @@ public static BrowserContext getOrCreateBrowserContext(ExtensionContext extensio
...
+    Tracing.StopOptions stopOptions = new Tracing.StopOptions().setPath(outputPath.resolve("trace.zip"));
+    browserContext.tracing().stop(stopOptions);
+  }
+
+  private static void createOutputPath(Path outputPath) {
+    try {
+      Files.walk(outputPath).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+      Files.createDirectories(outputPath);
+    } catch (NoSuchFileException ignored) {
+      // swallow
```
yury-s, line 160: This doesn't seem right, maybe wrap deletion into try catch but move createDirectories outside?

yury-s (approved):

yury-s:
> @yury-s was wondering if it's okay to land this feature without tests. The JUnit Engine Test Kit isn't playing well. I have confirmed that the issue is not in the feature but how the engine test kit runs the tests.

Yes, if it proves too difficult to test it, we can land it without tests. Would be nice to have some description of the problem here, so that later on if we wonder why it's difficult to write such tests we wouldn't need to repeat all the debugging and investigation that you've already done.

playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
uchagani: I was doing too much.  I think it's cleaned up now where if the directory exists, I don't touch it.

playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
```diff
@@ -66,10 +64,135 @@ public static BrowserContext getOrCreateBrowserContext(ExtensionContext extensio
...
+  }
+
+  private static void saveTraceWhenOn(ExtensionContext extensionContext) {
+    Options options = OptionsExtension.getOptions(extensionContext);
+    if (options.trace.equals(Options.Trace.ON)) {
+      saveTrace(extensionContext);
+    }
+  }
+
+  private static void saveScreenshot(ExtensionContext extensionContext, boolean didTestFail) {
```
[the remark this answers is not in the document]
yury-s: I'd remove it altogether and recommend tracing instead in all scenarios.

uchagani:
looking into failures

uchagani:
@yury-s All tests should be passing now

PR #1575, opened 2024-05-08
chore: roll 1.44.0

yury-s:


sivadatla:
@yury-s 
Though it is merged a weeks back, 1.44.0 is not released yet. Could you let us know by when we can expect the release?
Thank you!

PR #1627, opened 2024-07-16
fix: Replaced println with logMessage in DriverJar

T-Fowl:
Replaces System.out.println with logMessage in DriverJar so as to only print out log messages if the driver is in debug mode.

### DIFF
--- a/driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java
+++ b/driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java
@@ -74,7 +74,7 @@ public class DriverJar extends Driver {
       skip = System.getenv(PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD);
     }
     if (skip != null && !"0".equals(skip) && !"false".equals(skip)) {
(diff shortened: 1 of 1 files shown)


yury-s (commented):
We actually wanted to always have that message printed (regardless of whether `DEBUG=pw:install` is specified), this allows to avoid confusion when the variable is specified, but the browsers haven't been downloaded.

T-Fowl:
I can understand that - I wonder if it would be possible to print to stderr instead of stdout? This aligns with the unix [standard streams practice](https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html)

My personal use case is in CLI applications when using a remote browser for the automation, this debug message means I cannot pipe my output to other programs.

Originally I changed to `logMessage` because of the precedent set a few lines below when checking for a remote selenium url:

https://github.com/microsoft/playwright-java/blob/1e8adde480b50474b892992b34443a1fa7b9bab5/driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java#L80-L83

yury-s (approved):
Fair enough. Let's change it to logMessage. We will still print to stdout from `npx playwright install` when downloading browsers though, I think we may want to redirect that to logMessage too in java.

PR #1638, opened 2024-07-31
chore: roll driver to 1.46.0-beta

yury-s:
Fixes: https://github.com/microsoft/playwright-java/issues/1598


playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
```diff
@@ -412,4 +414,40 @@ static int fromJsRegexFlags(String regexFlags) {
...
+        }
+      } catch (IOException e) {
+        throw new PlaywrightException("Failed to read from file", e);
+      }
+      if (cert.passphrase != null) {
+        jsonCert.addProperty("passphrase", cert.passphrase);
+      }
+      clientCertificates.add(jsonCert);
+    }
+    params.remove("clientCertificates");
```
dgozman, line 450: Not that I mind, but it's surprising there could already be a property with this name.

playwright/src/test/java/com/microsoft/playwright/TestClientCertificates.java
```diff
@@ -0,0 +1,181 @@
...
+    super.stopServer();
+  }
+
+  @Test
+  public void shouldFailWithNoClientCertificatesProvided() {
+    APIRequestContext request = playwright.request().newContext(
+        new APIRequest.NewContextOptions().setIgnoreHTTPSErrors(true));
+    PlaywrightException e = assertThrows(PlaywrightException.class, () -> request.get(customServer.url));
+    assertTrue(e.getMessage().contains("Error: socket hang up"), e.getMessage());
+//    APIResponse response = request.get(customServer.url);
```
dgozman, line 63: Remove?

dgozman (approved):

playwright/src/test/java/com/microsoft/playwright/TestBrowserContextFetch.java
```diff
@@ -865,4 +865,28 @@ void shouldNotWorkAfterContextDispose() {
...
+        return;
+      }
+      exchange.getResponseHeaders().add("Content-Type", "text/plain");
+      exchange.sendResponseHeaders(200, 0);
+      try (OutputStreamWriter writer = new OutputStreamWriter(exchange.getResponseBody())) {
+        writer.write("Hello!");
+      }
+    });
+
+    APIRequestContext requestContext = context.request();
```
mxschmitt, line 884: this should use global fetch instead.

playwright/src/test/java/com/microsoft/playwright/TestGlobalFetch.java
```diff
@@ -523,4 +523,28 @@ void shouldSupportDisposeReason() {
...
+        return;
+      }
+      exchange.getResponseHeaders().add("Content-Type", "text/plain");
+      exchange.sendResponseHeaders(200, 0);
+      try (OutputStreamWriter writer = new OutputStreamWriter(exchange.getResponseBody())) {
+        writer.write("Hello!");
+      }
+    });
+
+    APIRequestContext requestContext = context.request();
```
mxschmitt, line 542: this should use global APIRequestContext instead.

playwright/src/test/java/com/microsoft/playwright/TestTap.java
```diff
@@ -97,6 +98,7 @@ void shouldNotSendMouseEventsWhenTouchendIsCanceled() {
   }
 
   @Test
+  @Disabled("noWaitAfter was removed from tap() in https://github.com/microsoft/playwright/pull/31739")
```
mxschmitt, line 101: In .NET I don't pass the values to the server anymore except for press/click/selectOption and adjusted the tests like we did upstream instead of disabling them.

mxschmitt (approved):

playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
yury-s: That's because we use `new Gson().toJSON(options)`, we can either nullify the option before the call or remove it from JSON like here.

PR #1649, opened 2024-08-12
devops: publish Ubuntu 24.04 Docker image

mxschmitt:


utils/docker/Dockerfile.noble
```diff
@@ -0,0 +1,53 @@
...
+    curl \
+    # Install utilities required for downloading driver
+    unzip \
+    # For the MSEdge install script
+    gpg && \
+    rm -rf /var/lib/apt/lists/* && \
+    # Create the pwuser
+    adduser pwuser
+
+# Ubuntu 22.04 and earlier come with Maven 3.6.3 which fails with
```
dgozman, line 21: Is this relevant to 24.04? Let's double check?

dgozman (approved):

utils/docker/Dockerfile.noble
yury-s: Let's check this

yury-s (approved):

utils/docker/Dockerfile.noble
mxschmitt: As per offline discussion with Yury:

> The maven version we distribute is never (3.8.7 vs. 3.9.x) than what users get in Ubuntu 24.04 via apt.
I think its better to have it manually installed like we do, then its aligned in all our 3 images, wdyt?

PR #1669, opened 2024-09-09
test: fix client-certificate tests

mxschmitt:
The problem before was that the Java WebServer was configured in a way, that when no client certificates were passed during the TLS handshake, it was closing the connection / aborting the handshake immediately. This meant for us that we were closing the Socks connections which triggered `is interrupted by another navigation` or `EMPTY_RESPONSE`. Also this lead to differences to the tests compared to upstream and failing tests across the board.

This PR aligns it with upstream and fixed the tests.

### DIFF
--- a/playwright/src/test/java/com/microsoft/playwright/ServerWithClientCertificate.java
+++ b/playwright/src/test/java/com/microsoft/playwright/ServerWithClientCertificate.java
@@ -136,8 +136,8 @@ public class ServerWithClientCertificate implements HttpHandler {
   public void handle(HttpExchange exchange) throws IOException {
     SSLSession sslSession = ((HttpsExchange) exchange).getSSLSession();
     String response = div("servername", sslSession.getPeerHost());
-    Certificate[] certs = sslSession.getPeerCertificates();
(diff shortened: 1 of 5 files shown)


playwright/src/test/java/com/microsoft/playwright/ServerWithClientCertificate.java
```diff
@@ -149,7 +149,7 @@ public void handle(HttpExchange exchange) throws IOException {
           cert.getSubjectX500Principal().getName(), cert.getIssuerX500Principal().getName()));
         exchange.sendResponseHeaders(403, 0);
       }
-    } else {
+    } catch (SSLPeerUnverifiedException e) {
```
yury-s, line 152: What about NPE and index out of range exceptions which may happen if there  are no certs?

yury-s (approved):

playwright/src/test/java/com/microsoft/playwright/ServerWithClientCertificate.java
mxschmitt: happy to add it back but for some reason it wasn't used, so I thought removing it is better then.

PR #1681, opened 2024-10-02
chore: roll 1.48 beta driver

yury-s:
* Roll driver
* Implement routeWebSocket

Fixes https://github.com/microsoft/playwright-java/issues/1666


playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
```diff
@@ -457,4 +457,29 @@ private static String base64Buffer(byte[] bytes, Path path) throws IOException {
...
+      Object urlFilter = matcher.rawSource;
+      if (urlFilter instanceof String) {
+        jsonPattern.addProperty("glob", (String) urlFilter);
+      } else if (urlFilter instanceof Pattern) {
+        Pattern pattern = (Pattern) urlFilter;
+        jsonPattern.addProperty("regexSource", pattern.pattern());
+        jsonPattern.addProperty("regexFlags", toJsRegexFlags(pattern));
+      } else {
+        // Match all requests.
+        jsonPattern.addProperty("glob", "**/*");
```
dgozman, line 474: Do you want to remove all other properties added prior to this point? It should not matter technically, but would be cleaner.

playwright/src/test/java/com/microsoft/playwright/TestRouteWebSocket.java
```diff
@@ -0,0 +1,180 @@
...
+  @ValueSource(strings = {"no-mock", "no-match", "pass-through"})
+  public void shouldWorkWithTextMessage(String mock, Page page) throws Exception {
+    setupRoute(page, mock);
+    Future<WebSocket> wsPromise = webSocketServer.waitForWebSocket();
+    setupWS(page, webSocketServer.getPort(), "blob");
+
+    page.waitForCondition(() -> {
+      Boolean result = (Boolean) page.evaluate("() => window.log.length >= 1");
+      return result;
+    }, new Page.WaitForConditionOptions().setTimeout(10_000));
```
dgozman, line 103: Is there a way to set default timeout for `waitForCondition`? If not, we should consider adding that.

playwright/src/test/java/com/microsoft/playwright/TestRouteWebSocket.java
```diff
@@ -0,0 +1,180 @@
...
+    assertEquals(
+      asList("open", "message: data=arraybuffer:hi origin=ws://localhost:" + webSocketServer.getPort() + " lastEventId="),
+      page.evaluate("window.log"));
+    Future<String> messagePromise = webSocketServer.waitForMessage();
+    page.evaluate("() => window.ws.send(new Blob([new Uint8Array(['h'.charCodeAt(0), 'i'.charCodeAt(0)])]))");
+    // Without this the blob message is not sent in pass-through!
+    assertEquals(1, page.evaluate("window.ws.readyState"));
+    assertEquals("hi", messagePromise.get());
+  }
+}
```
dgozman, line 180: I think it would be valuable to also port "should work with server" and/or "should work without server", because these tests actually exercise all of the client APIs.

dgozman (approved):

playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
yury-s: There should be no properties as the jsonPatter was just created at line 464.

playwright/src/test/java/com/microsoft/playwright/TestRouteWebSocket.java
yury-s: It respects default timeout, I was setting it to lower value for local testing, removed.

playwright/src/test/java/com/microsoft/playwright/TestRouteWebSocket.java
yury-s: added

PR #1692, opened 2024-10-27
junit: allow user to reset Browser after each test

### LINKED ISSUE #1691: [Feature]: JUnit:  Allow the user to control the Browser object's lifecycle
### 🚀 Feature Request

Create a public method that will allow the user to close the browser and remove the reference from the respective `ThreadLocal` object.  

### Implementation 

```java
public static void closeBrowser() {
  Browser browser = threadLocalBrowser.get();
  if(browser == null) {
    return;
  }
  
  browser.close();
  threadLocalBrowser.remove();
}
```

### Usage

```java
@AfterEach
public void closeBrowser() {
  if(Boolean.parseBoolean(System.getProperty("browserStack"))) {
    BrowserExtension.closeBrowser();
  }
}
```

After the above `AfterEach` hook runs in the user's project, when the next test runs, the Browser will be re-created by the fixture using the existing options.  This will allow a new test session to start in BrowserStack.

### Motivation

Cloud testing services like BrowserStack launch new sessions using `BrowserType.connect`.  Each call to `connect` starts a new test session.

This doesn't work properly with current JUnit fixtures because the Browser object is re-used per thread and so all tests that run on the same thread use the same test session in BrowserStack.  This becomes one giant test according to BrowserStack.

My suggestion would allow the user to decide when to re-use the browser, which works great locally, but not so well when using cloud testing services.

uchagani:
Resolves #1691

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
@@ -95,7 +95,7 @@ public class BrowserContextExtension implements ParameterResolver, TestWatcher {
 
   private void cleanup(Options options) {
     closeBrowserContext();
-    if(options.closeBrowserAfterEachTest) {
+    if(options.closeBrowserAfterEachTest != null && options.closeBrowserAfterEachTest) {
       BrowserExtension.closeBrowser();
     }
   }
--- a/playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
@@ -75,13 +75,13 @@ public class BrowserContextExtension implements ParameterResolver, TestWatcher {
   @Override
   public void testSuccessful(ExtensionContext extensionContext) {
     saveTraceWhenOn(extensionContext);
-    closeBrowserContext();
+    cleanup(OptionsExtension.getOptions(extensionContext));
   }
 
   @Override
   public void testAborted(ExtensionContext extensionContext, Throwable cause) {
     saveTraceWhenOn(extensionContext);
-    closeBrowserContext();
+    cleanup(OptionsExtension.getOptions(extensionContext));
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/junit/BrowserContextExtension.java
```diff
@@ -90,7 +90,14 @@ public void testFailed(ExtensionContext extensionContext, Throwable cause) {
     if (shouldRecordTrace(options)) {
       saveTrace(extensionContext);
     }
+    cleanup(options);
+  }
+
+  private void cleanup(Options options) {
     closeBrowserContext();
+    if(options.closeBrowserAfterEachTest != null && options.closeBrowserAfterEachTest) {
+      BrowserExtension.closeBrowser();
```
yury-s, line 99: The behavior of closing the browser (rather than just the context) after each test is very specific to BrowserStack implementation and they could solve it better by providing an API for explicitly marking test start/end instead of binding it to the lifetime of the browser connection.

I'd rather the clients, that need to close browsers after each test, implemented their own fixture and just called `browser.close()` after each test from the fixture. We can add [`onDisconnect`](https://playwright.dev/java/docs/api/class-browser#browser-event-disconnected) handler to all browser instances that are created via `Playwright` fixture and clean-up thread local cache when the event is fired. It would also cover the case when the user calls `browser.close()` in the middle of a test too, so it's something that would make sense to do in any case.

Do you think such approach would make more sense or you'd still prefer to have the option?

playwright/src/main/java/com/microsoft/playwright/junit/Options.java
```diff
@@ -50,13 +50,21 @@ public class Options {
...
+  // Usually only necessary when using a cloud testing service
+  public Boolean closeBrowserAfterEachTest;
 
   public enum Trace {
     OFF,
     ON,
     RETAIN_ON_FAILURE;
   }
 
+  public Options setCloseBrowserAfterEachTest(Boolean closeBrowser) {
```
yury-s, line 63: If we proceed with this option, let's have a test for it.

uchagani:
going to close this PR and open a new one with the changes we discussed in #1691

Issue #1715, opened 2024-12-17
[Bug]: Allow WebSocketRoute matching without trailing slash


mxschmitt:
Reference: https://github.com/microsoft/playwright-python/issues/2686

PR #1717, opened 2024-12-18
chore: follow upstream url matcher

mxschmitt:
Fixes https://github.com/microsoft/playwright-java/issues/1715

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
@@ -16,34 +16,29 @@
 
 package com.microsoft.playwright.impl;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonObject;
 import com.microsoft.playwright.PlaywrightException;
 
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.Objects;
 import java.util.function.Predicate;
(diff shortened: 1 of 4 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -16,34 +16,29 @@
...
-  }
+  private final URL baseURL;
+  public final String glob;
+  public final Pattern pattern;
+  public final Predicate<String> predicate;
 
   static UrlMatcher forOneOf(URL baseUrl, Object object) {
     if (object == null) {
-      return UrlMatcher.any();
+      return new UrlMatcher(baseUrl, (String) null);
```
yury-s, line 41: This will only match baseURL while it was matching anything before.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -16,34 +16,29 @@
...
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
+import java.util.ArrayList;
+import java.util.List;
 
 import static com.microsoft.playwright.impl.Utils.globToRegex;
 import static com.microsoft.playwright.impl.Utils.toJsRegexFlags;
 
 class UrlMatcher {
-  final Object rawSource;
```
yury-s, line 33: Let's revert this part, it can be refactored in a separate change.

PR #1720, opened 2024-12-19
chore: refactor UrlMatcher

mxschmitt:
Extracted from https://github.com/microsoft/playwright-java/pull/1717. There should be no behaviour changes.

This moves `rawSource` towards having it lazily determined which matching source it is - this helps us to align with upstream.

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
@@ -84,10 +84,6 @@ class UrlMatcher {
   }
 
   boolean test(String value) {
-    return testImpl(baseURL, pattern, predicate, glob, value);
-  }
-
-  private static boolean testImpl(URL baseURL, Pattern pattern, Predicate<String> predicate, String glob, String value) {
     if (pattern != null) {
       return pattern.matcher(value).find();
     }
@@ -105,14 +101,14 @@ class UrlMatcher {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     UrlMatcher that = (UrlMatcher) o;
-    if (pattern != null && !pattern.pattern().equals(that.pattern.pattern()) && pattern.flags() == that.pattern.flags()) {
-      return false;
+    if (pattern != null) {
+      return that.pattern != null && pattern.pattern().equals(that.pattern.pattern()) && pattern.flags() == that.pattern.flags();
     }
-    if (predicate != null && !predicate.equals(that.predicate)) {
-      return false;
+    if (predicate != null) {
+      return predicate.equals(that.predicate);
     }
-    if (glob != null && !glob.equals(that.glob)) {
-      return false;
+    if (glob != null) {
+      return glob.equals(that.glob);
     }
     return true;
   }
@@ -125,7 +121,10 @@ class UrlMatcher {
     if (predicate != null) {
       return predicate.hashCode();
     }
-    return glob.hashCode();
+    if (glob != null) {
+      return glob.hashCode();
+    }
+    return super.hashCode();
   }
 
   @Override
--- a/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
@@ -110,7 +110,7 @@ class UrlMatcher {
     if (glob != null) {
       return glob.equals(that.glob);
     }
-    return true;
+    return that.pattern == null  && that.predicate == null && that.glob == null;
   }
 
   @Override
--- a/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
@@ -124,7 +124,7 @@ class UrlMatcher {
     if (glob != null) {
       return glob.hashCode();
     }
-    return super.hashCode();
+    return 0;
   }
 
   @Override
--- a/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
@@ -22,6 +22,8 @@ import com.microsoft.playwright.PlaywrightException;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,79 @@ static String resolveUrl(URL baseUrl, String spec) {
...
+    this.pattern = pattern;
     this.predicate = predicate;
   }
 
   boolean test(String value) {
-    return predicate == null || predicate.test(value);
+    return testImpl(baseURL, pattern, predicate, glob, value);
+  }
+
+  private static boolean testImpl(URL baseURL, Pattern pattern, Predicate<String> predicate, String glob, String value) {
```
mxschmitt, line 89: why another method? We'll later on (in future patches) mutate e.g. the `glob` env var - making it `private static` helps us there not to rewrite the UrlMatcher's instance property itself.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -30,20 +32,14 @@
...
-  }
+  private final URL baseURL;
+  public final String glob;
+  public final Pattern pattern;
+  public final Predicate<String> predicate;
 
   static UrlMatcher forOneOf(URL baseUrl, Object object) {
     if (object == null) {
-      return UrlMatcher.any();
+      return new UrlMatcher((Predicate<String>) s -> true);
```
yury-s, line 42: Now it is indistinguishable from an actual predicate in toString.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,79 @@ static String resolveUrl(URL baseUrl, String spec) {
...
   public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     UrlMatcher that = (UrlMatcher) o;
-    if (rawSource instanceof Pattern && that.rawSource instanceof Pattern) {
-      Pattern a = (Pattern) rawSource;
-      Pattern b = (Pattern) that.rawSource;
-      return a.pattern().equals(b.pattern()) && a.flags() == b.flags();
+    List<Boolean> matches = new ArrayList<>();
+    if (baseURL != null && that.baseURL != null) {
```
yury-s, line 108: I don't think we care if the pattern is a result of a base url + glob or just empty base url and a glob as long as the final glob is the same.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -30,20 +32,14 @@
...
-  private final Predicate<String> predicate;
-
-  private static Predicate<String> toPredicate(Pattern pattern) {
-    return s -> pattern.matcher(s).find();
-  }
-
-  static UrlMatcher any() {
-    return new UrlMatcher((Object) null, null);
-  }
+  private final URL baseURL;
```
yury-s, line 35: Why not resolve the pattern in the constructor and avoid plumbing base url all the way through?

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,79 @@ static String resolveUrl(URL baseUrl, String spec) {
...
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     UrlMatcher that = (UrlMatcher) o;
-    if (rawSource instanceof Pattern && that.rawSource instanceof Pattern) {
-      Pattern a = (Pattern) rawSource;
-      Pattern b = (Pattern) that.rawSource;
-      return a.pattern().equals(b.pattern()) && a.flags() == b.flags();
+    List<Boolean> matches = new ArrayList<>();
+    if (baseURL != null && that.baseURL != null) {
+      matches.add(baseURL.equals(that.baseURL));
```
yury-s, line 109: Please use early return here and in other places.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
mxschmitt: We need this later on when I'll add this check:

https://github.com/microsoft/playwright/blob/edd789780ac473083f6fb0b8740faddc43b439ab/packages/playwright-core/src/utils/isomorphic/urlMatch.ts#L101-L103

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
mxschmitt: Better like that?

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,72 @@ static String resolveUrl(URL baseUrl, String spec) {
...
   @Override
   public int hashCode() {
-    return Objects.hash(rawSource);
+    if (pattern != null) {
+      return pattern.hashCode();
+    }
+    if (predicate != null) {
+      return predicate.hashCode();
+    }
+    return glob.hashCode();
```
yury-s, line 124: `glob` can also be `null`.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,72 @@ static String resolveUrl(URL baseUrl, String spec) {
...
+      return that.pattern != null && pattern.pattern().equals(that.pattern.pattern()) && pattern.flags() == that.pattern.flags();
+    }
+    if (predicate != null) {
+      return predicate.equals(that.predicate);
     }
-    return Objects.equals(rawSource, that.rawSource);
+    if (glob != null) {
+      return glob.equals(that.glob);
+    }
+    return false;
```
yury-s, line 113: I think `<any>` is equal to `<any>`, isn't it?

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,75 @@ static String resolveUrl(URL baseUrl, String spec) {
...
+      return that.pattern != null && pattern.pattern().equals(that.pattern.pattern()) && pattern.flags() == that.pattern.flags();
+    }
+    if (predicate != null) {
+      return predicate.equals(that.predicate);
+    }
+    if (glob != null) {
+      return glob.equals(that.glob);
     }
-    return Objects.equals(rawSource, that.rawSource);
+    return true;
```
yury-s, line 113: I believe it should be `return that.pattern == null  && that.predicate == null && that.glob == null;`, otherwise it will be equal to any other pattern.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -68,50 +64,75 @@ static String resolveUrl(URL baseUrl, String spec) {
...
+    if (pattern != null) {
+      return pattern.hashCode();
+    }
+    if (predicate != null) {
+      return predicate.hashCode();
+    }
+    if (glob != null) {
+      return glob.hashCode();
+    }
+    return super.hashCode();
```
yury-s, line 127: This will vary depending on the object instance and we want it to be a constant for `<any>` pattern, `return 1` or something?

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
mxschmitt: I think I already have it like that.

yury-s: I disagree, with the current code e.g.`new UrlMatcher(null, null, null, null).equals(new UrlMatcher((String s) -> true) == true` which is incorrect.

yury-s (approved):

PR #1722, opened 2024-12-28
fix(webSocketRoute): resolve URL against baseURL

mxschmitt:
Note: `URL` is not able to parse a `ws://` url, only `URI` is able to.

This applies [this](https://github.com/microsoft/playwright/blob/4819747c85faa8cad0f33d770e1e63f6fb894d49/packages/playwright-core/src/utils/isomorphic/urlMatch.ts#L101-L103) upstream logic to Java.

Relates https://github.com/microsoft/playwright-java/issues/1715
Extracted from https://github.com/microsoft/playwright-java/pull/1717

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
@@ -16,15 +16,11 @@
 
 package com.microsoft.playwright.impl;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonObject;
 import com.microsoft.playwright.PlaywrightException;
 
-import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -77,21 +78,32 @@ static String resolveUrl(URL baseUrl, String spec) {
...
     }
     if (predicate != null) {
       return predicate.test(value);
     }
     if (glob != null) {
-      return Pattern.compile(globToRegex(resolveUrl(baseURL, glob))).matcher(value).find();
+      if (!glob.startsWith("*")) {
+        // Allow http(s) baseURL to match ws(s) urls.
+        if (baseURL != null && Pattern.compile("^https?://").matcher(baseURL).find() && Pattern.compile("^wss?://").matcher(value).find()) {
+          baseURL = baseURL.replaceFirst("^https?", "ws");
```
yury-s, line 102: This will replace both `http` and `https` with `ws`. I assume we want to replace https->wss and http->ws, don't we?

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -77,21 +78,32 @@ static String resolveUrl(URL baseUrl, String spec) {
...
     if (pattern != null) {
       return pattern.matcher(value).find();
     }
     if (predicate != null) {
       return predicate.test(value);
     }
     if (glob != null) {
-      return Pattern.compile(globToRegex(resolveUrl(baseURL, glob))).matcher(value).find();
+      if (!glob.startsWith("*")) {
+        // Allow http(s) baseURL to match ws(s) urls.
```
mxschmitt, line 100: This mimics https://github.com/microsoft/playwright/blob/640e6a8aa75a23a31e7591f34176a32ed73ba022/packages/playwright-core/src/utils/isomorphic/urlMatch.ts#L101-L103

dgozman (approved):

PR #1746, opened 2025-02-07
fix: add pfxBase64 to jsonCert instead of params

stevenfuhr:
This patch fixes PKCS12 client certificates not being added to the context. client_keystore is created from cert.pem and key.pem with password "passphrase".

Fixes https://github.com/microsoft/playwright-java/issues/1747


stevenfuhr:

yury-s (approved):
Good catch!

Issue #1736, opened 2025-01-29
Dependency is outdated


EugenFischer:
Please update the playwright dependency version in examples/pom.xml [<version>1.41.0</version>](https://github.com/microsoft/playwright-java/blob/48200884578e8a95d807b2f2beb63ab41351aaa6/examples/pom.xml#L18)

Kind Regards
Eugen Fischer

PR #1748, opened 2025-02-10
chore: auto update example pw version in pom

mxschmitt:
Fixes https://github.com/microsoft/playwright-java/issues/1736

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -10,13 +10,13 @@
   <name>Playwright Client Examples</name>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <playwright-version>1.49.0</playwright-version>
+    <playwright.version>1.49.0</playwright.version>
   </properties>
   <dependencies>
     <dependency>
       <groupId>com.microsoft.playwright</groupId>
       <artifactId>playwright</artifactId>
-      <version>${playwright-version}</version>
+      <version>${playwright.version}</version>
     </dependency>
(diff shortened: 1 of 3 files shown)


examples/pom.xml
```diff
@@ -10,12 +10,13 @@
   <name>Playwright Client Examples</name>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <playwright-version>1.49.0</playwright-version>
```
yury-s, line 13:
```suggestion
    <playwright.version>1.49.0</playwright.version>
```

scripts/set_maven_version.sh
```diff
@@ -20,7 +20,12 @@ POM_FILES=(
...
 )
 
-for name in ${POM_FILES[*]};
-do
-  mvn versions:set -D generateBackupPoms=false -D newVersion=$VERSION -f $name
-done
+node -e "$(cat <<EOF
+  const parts = process.argv[1].split('.').map(part => parseInt(part, 10));
+  parts[1]--;
+  const previousMajorVersion = parts.join('.');
```
yury-s, line 26: This is not gonna work, on the `main` branch the version is `1.X.Y-SNAPSHOT`, we need to drop the -snapshot part there.

yury-s (approved):
Btw, it only makes sense to update the version on the `main` branch, where our examples are loaded from. We don't run `set_maven_version.sh` regularly on the main branch today.

PR #1760, opened 2025-03-14
chore: update readme playwright dependency version

lubomsky:


tools/update-docs-version/src/main/java/com/microsoft/playwright/tools/UpdatePlaywrightVersion.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class UpdatePlaywrightVersion {
+  private static final Pattern VERSION_PATTERN = Pattern.compile(
+    "<groupId>com.microsoft.playwright</groupId>\\s*<artifactId>parent-pom</artifactId>\\s*<version>(\\d+\\.\\d+\\.\\d+)(?:-SNAPSHOT)?</version>"
+  );
+  private static final Pattern MAVEN_PATTERN   = Pattern.compile(
+    "(<dependency>\\s*<groupId>com.microsoft.playwright</groupId>\\s*<artifactId>playwright</artifactId>\\s*)<version>\\d+\\.\\d+\\.\\d+</version>(\\s*</dependency>)"
+  );
```
yury-s, line 19: Does it much multiline by default (without MULTILINE flag) ?

tools/update-docs-version/src/main/java/com/microsoft/playwright/tools/UpdatePlaywrightVersion.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public static void main(String[] args) throws Exception {
+    Path   pomPath    = Paths.get("pom.xml");
+    String pomContent = Files.readString(pomPath, UTF_8);
+
+    Matcher versionMatcher = VERSION_PATTERN.matcher(pomContent);
+    if (!versionMatcher.find()) {
+      throw new NoSuchElementException("Project version was not found");
+    }
+    String version = versionMatcher.group(1);
```
yury-s, line 32: Version in pom.xml usually refers to the next development version, which hasn't been published yet. Let's set it to major.(minor-1).0 similar to what we did in https://github.com/microsoft/playwright-java/pull/1748.

Also would be nice to unify this two places, sot that the versions are patched at the same place with the same deps version.

tools/update-docs-version/src/main/java/com/microsoft/playwright/tools/UpdatePlaywrightVersion.java
lubomsky: Yes
Zrzut ekranu 2025-03-14 o 18 03 47

yury-s (commented):
After discussing this offline, we decided to align the readme file with our other language ports and redirect to the main documentation site playwright.dev. We already have infra that keeps the versions in the examples up to date, and it's mote comprehensive than this README. After https://github.com/microsoft/playwright-java/pull/1766 there will be no mentions of package versions in the readme.

yury-s:
Closing per the comment above. There is no version in the readme any more.

PR #1761, opened 2025-03-15
Test/warning cleanups

mxschmitt:
- GHA Timeouts were increased to 60min
- Verified that retries are working, weird that they didn't seem to work on CI
- Introduced `PW_JAVA_INTERNAL_SILENT_INSTALL` to have no confusing output in the test output - Overriding the stderr/out from inside the test does not work, since ProcessBuilder is based on file descriptors.
- Had to add `slf4j-simple` in order to get rid of:
```
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
```

tbd. Lets see if increasing the timeouts helps to make the tests pass.


.github/workflows/test.yml
```diff
@@ -12,17 +12,17 @@ env:
   PW_MAX_RETRIES: 3
 jobs:
   dev:
-    timeout-minutes: 30
+    timeout-minutes: 60
```
yury-s, line 15: Let's do it only for channels that we know are slow?

playwright/pom.xml
```diff
@@ -57,6 +57,10 @@
       <groupId>org.java-websocket</groupId>
       <artifactId>Java-WebSocket</artifactId>
     </dependency>
+    <dependency>
```
yury-s, line 60: let's remove?

pom.xml
```diff
@@ -91,6 +92,12 @@
         <version>${websocket.version}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
```
yury-s, line 96: Let's add comment that it's just for websocket to stop polluting the output with warnings.

pom.xml
```diff
@@ -151,7 +158,7 @@
           <configuration>
             <properties>
               <configurationParameters>
-                junit.jupiter.execution.parallel.enabled = true
+                junit.jupiter.execution.parallel.enabled = false
```
yury-s, line 161: stray change?

driver-bundle/src/test/java/com/microsoft/playwright/impl/driver/jar/TestInstall.java
```diff
@@ -74,6 +74,7 @@ void shouldThrowWhenBrowserPathIsInvalid(@TempDir Path tmpDir) throws NoSuchFiel
     // Make sure the browsers are not installed yet by pointing at an empty dir.
     env.put("PLAYWRIGHT_BROWSERS_PATH", tmpDir.toString());
     env.put("PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD", "false");
+    env.put("PW_JAVA_INTERNAL_SILENT_INSTALL", "true");
```
yury-s, line 77: Let's not do it for now. If we do it, let's make it a public option?

yury-s (approved):

PR #1762, opened 2025-03-16
devops: add DevContainer config

mxschmitt:
Otherwise it defaults to 'https://github.com/devcontainers/images/tree/main/src/universal' which is Ubuntu 20.04.


.github/dependabot.yml
yury-s, line 1: Let's revert this file.

yury-s (approved):

PR #1780, opened 2025-04-24
chore: roll 1.52.0 driver, implement new features

Skn0tt:
Closes https://github.com/microsoft/playwright-java/issues/1753.

I don't think it's worth it to implement an entire proxy for https://github.com/microsoft/playwright/pull/35389, so I didn't add a test for that.


playwright/src/test/java/com/microsoft/playwright/TestGlobalFetch.java
```diff
@@ -370,6 +371,26 @@ void shouldThrowAnErrorWhenMaxRedirectsIsExceeded() {
     request.dispose();
   }
 
+  @Test
+  void shouldThrowAnErrorWhenMaxRedirectsOnContextIsExceeded() {
```
yury-s, line 375: Let's preserve the test name and port other tests from https://github.com/microsoft/playwright/pull/35160

playwright/src/main/java/com/microsoft/playwright/impl/Utils.java
```diff
@@ -91,79 +89,6 @@ public static <T> T clone(T f) {
     return convertType(f, (Class<T>) f.getClass());
   }
 
-
-  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#escaping
```
yury-s, line 95: Nice!

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -66,61 +63,36 @@ private static String resolveUrl(String baseUrl, String spec) {
...
-    this.baseURL = baseURL != null ? baseURL.toString() : null;
+  private UrlMatcher(String glob, Pattern pattern, Predicate<String> predicate) {
     this.glob = glob;
     this.pattern = pattern;
     this.predicate = predicate;
   }
 
   boolean test(String value) {
-    return testImpl(baseURL, pattern, predicate, glob, value);
+    return testImpl(pattern, predicate, glob, value);
```
yury-s, line 86: No need to pass glob anymore.

playwright/src/main/java/com/microsoft/playwright/impl/UrlMatcher.java
```diff
@@ -66,61 +63,36 @@ private static String resolveUrl(String baseUrl, String spec) {
...
   }
 
-  private static boolean testImpl(String baseURL, Pattern pattern, Predicate<String> predicate, String glob, String value) {
+  private static boolean testImpl(Pattern pattern, Predicate<String> predicate, String glob, String value) {
```
yury-s, line 89: Just inline this method, it has only one caller now.

playwright/src/test/java/com/microsoft/playwright/TestPageInterception.java
```diff
@@ -139,15 +139,4 @@ void shouldNotFollowRedirectsWhenMaxRedirectsIsSetTo0InRouteFetch() {
     page.navigate(server.PREFIX + "/foo");
     assertTrue(page.content().contains("hello"));
   }
-
-  @Test
-  void shouldProperlyHandleCharacterSetsInGlobs() {
```
yury-s, line 144: Let's port 'should work with glob' from upstream, to make sure our logic doesn't break when constructing java regex from the node's string.

playwright/src/test/java/com/microsoft/playwright/TestPageAriaSnapshot.java
```diff
@@ -67,7 +70,13 @@ void shouldSnapshotListWithAccessibleName(Page page) {
   @Test
   void shouldSnapshotComplex(Page page) {
     page.setContent("<ul><li><a href='about:blank'>link</a></li></ul>");
-    checkAndMatchSnapshot(page.locator("body"), "- list:\n  - listitem:\n    - link \"link\"");
+    checkAndMatchSnapshot(page.locator("body"), "- list:\n  - listitem:\n    - link \"link\":\n      - /url: about:blank");
```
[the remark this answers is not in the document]
Skn0tt: none of those changed tests exist in this file - am I looking at it wrong?

Skn0tt:
> Looks good, but make sure that all the upstream tests are ported.

I had LLM do the work for me. This was the prompt:

```md
This project is a port of the Typescript version of Playwright. I'm updating it, and it's important we also update tests. I'll give you a diff of the Typescript tests. Please apply the same changes to the tests in this project, if we have corresponding tests.

<git diff release-1.51 release-1.52  -- tests/page tests/library>
```

Do we need to update anything else?

playwright/src/test/java/com/microsoft/playwright/TestPageAriaSnapshot.java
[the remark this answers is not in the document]
yury-s: They were probably not ported before. Not all the tests were translated originally when the java port was introduced. But over the last few years we've been trying to port at least all the tests related to the upstream changes when rolling them to java (and other language ports). `interception.spec.ts` covers a lot of important edge cases related to glob->regex transformation and it is important to have them in java as well, so that we can be sure that the local utils approach works.

playwright/src/test/java/com/microsoft/playwright/TestPageInterception.java
```diff
@@ -139,4 +142,93 @@ void shouldNotFollowRedirectsWhenMaxRedirectsIsSetTo0InRouteFetch() {
...
+
+  Pattern globToRegex(String glob) {
+    return globToRegex(glob, null, false);
+  }
+
+  Pattern globToRegex(String glob, String baseURL, boolean webSocketUrl) {
+    return ((PlaywrightImpl) playwright).localUtils().globToRegex(glob, baseURL, webSocketUrl);
+  }
+
+  boolean urlMatches(String baseURL, String urlString, Object match) {
```
yury-s, line 213: I can only see calls with 3 string parameters, why is match an object?

playwright/src/test/java/com/microsoft/playwright/TestPageInterception.java
```diff
@@ -139,4 +142,93 @@ void shouldNotFollowRedirectsWhenMaxRedirectsIsSetTo0InRouteFetch() {
...
+    return urlMatches(baseURL, urlString, match, false);
+  }
+
+  boolean urlMatches(String baseURL, String urlString, Object match, boolean webSocketUrl) {
```
yury-s, line 217: Please remove unused signatures and code, looks like the LLM didn't clean it up.

yury-s (approved):

yury-s (approved):

PR #1807, opened 2025-06-20
chore: require explicit timeout parameter in sendMessage()

yury-s:
Either `NO_TIMEOUT` or computed timeout must be provided for each call to `sendMessage(method, params, timeout)`, so that we can leverage compiler support.

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/APIRequestImpl.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/APIRequestImpl.java
@@ -27,6 +27,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.List;
 
+import static com.microsoft.playwright.impl.ChannelOwner.NO_TIMEOUT;
 import static com.microsoft.playwright.impl.Serialization.gson;
 import static com.microsoft.playwright.impl.Utils.addToProtocol;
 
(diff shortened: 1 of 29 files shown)


playwright/src/main/java/com/microsoft/playwright/impl/APIRequestImpl.java
```diff
@@ -65,7 +66,8 @@ public APIRequestContextImpl newContext(NewContextOptions options) {
       params.add("storageState", storageState);
     }
     addToProtocol(params, clientCertificateList);
-    JsonObject result = playwright.sendMessage("newRequest", params).getAsJsonObject();
+    // Timeout here is not the command's timeout, but we pass it for consistency.
+    JsonObject result = playwright.sendMessage("newRequest", params, options.timeout).getAsJsonObject();
```
dgozman, line 70: Let's pass `NO_TIMEOUT` instead? Otherwise we might get unexpected timeouts later on, when server starts supporting the timeout here.

dgozman (approved):

playwright/src/main/java/com/microsoft/playwright/impl/APIRequestImpl.java
yury-s: Ok, fixed as discussed.

mxschmitt (approved):

Issue #1447, opened 2024-01-11
[BUG] Failed to create driver when using in sbt plugin


gmkumar2005:
### System info
- Playwright Version: [1.40.0]
- Operating System: [macOS] (Not tested on other os. It is fair to assume this issue exists on all OS)
- Browser: [All]
- Other info:

**Expected**
The playwright drivers should be loaded and open the browser when running in an sbt plugin. 

I am using playwright in an sbt plugin. It fails to load the driver with following error 
`Caused by: java.lang.NullPointerException: Cannot invoke "java.net.URL.toURI()" because the return value of "java.lang.ClassLoader.getResource(String)" is null`

**Actual**

I am using playwright in an sbt plugin. It fails to load the driver with following error 
`Caused by: java.lang.NullPointerException: Cannot invoke "java.net.URL.toURI()" because the return value of "java.lang.ClassLoader.getResource(String)" is null`

The root cause seems to be due to the way class loader is used in the `DriverJar.Java`
The function `getDriverResourceURI()` is using a classloader from current Thread 
`ClassLoader classloader = Thread.currentThread().getContextClassLoader(); `

sbt is not able to provide all the libraries inthis class loader. It works with `DriverJar.class.getClassLoader() `

**Can this function `getDriverResourceURI()` be updated to use the class loader as in DriverJar.class.getClassLoader()**

https://github.com/microsoft/playwright-java/blob/e889b20fda31695d0815cdc1eb3cde7efd54db73/driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java#L117

yury-s:
Sounds like a problem with sbt as it works in regular environment. Providing more context on how you use Playwright and how this can be reproduced, ideally with a minimal repro that we could run locally, would help us better understand what's going on. I don't know what sbt is, but it sounds like some environment that would be out of scope for the project. With that said, if you have an idea for the fix we accept contributions.

gmkumar2005:
I fixed this issue by writing a custom DriverJar. I changed the classLoading logic to `DriverJar.class.getClassLoader() ` .
To make it work system property is setup as below.

`System.setProperty("playwright.driver.impl", "jsenv.DriverJar")`

gmkumar2005:
Issue is fixed please see previous comment

PR #1811, opened 2025-06-24
fix: Replaced classLoader in DriverJar

### LINKED ISSUE #1803: [Bug]: java.lang.RuntimeException: Failed to create driver
### Version

v1.50.0-jammy

### Steps to reproduce

Image

### Expected behavior

Normal navigation webpage

### Actual behavior

java.lang.RuntimeException: Failed to create driver
        at com.microsoft.playwright.impl.driver.Driver.createAndInstall(Driver.java:105) ~[driver-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.driver.Driver.ensureDriverInstalled(Driver.java:57) ~[driver-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.PlaywrightImpl.createImpl(PlaywrightImpl.java:46) ~[playwright-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.PlaywrightImpl.create(PlaywrightImpl.java:36) ~[playwright-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.Playwright.create(Playwright.java:110) ~[playwright-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.Playwright.create(Playwright.java:114) ~[playwright-1.50.0.jar!/:1.50.0]
        at com.toowe.report.common.web.spider.WebSpiderUtil.spiderUrlRequest(WebSpiderUtil.java:119) ~[report-common-0.0.1.jar!/:0.0.1]
        at com.toowe.report.common.tools.web.spider.WebUrlSpider.spiderUrlRequest(WebUrlSpider.java:18) ~[report-common-0.0.1.jar!/:0.0.1]
        at com.toowe.report.common.tools.web.search.BingWebSearchEngine.lambda$search$0(BingWebSearchEngine.java:94) ~[report-common-0.0.1.jar!/:0.0.1]
        at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796) ~[na:na]
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387) ~[na:na]
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312) ~[na:na]
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843) ~[na:na]
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808) ~[na:na]
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188) ~[na:na]
Caused by: java.lang.NullPointerException: Cannot invoke "java.net.URL.toURI()" because the return value of "java.lang.ClassLoader.getResource(String)" is null
        at com.microsoft.playwright.impl.driver.jar.DriverJar.getDriverResourceURI(DriverJar.java:118) ~[driver-bundle-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.driver.jar.DriverJar.extractDriverToTempDir(DriverJar.java:122) ~[driver-bundle-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.driver.jar.DriverJar.initialize(DriverJar.java:65) ~[driver-bundle-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.driver.Driver.initialize(Driver.java:64) ~[driver-1.50.0.jar!/:1.50.0]
        at com.microsoft.playwright.impl.driver.Driver.createAndInstall(Driver.java:101) ~[driver-1.50.0.jar!/:1.50.0]
        ... 15 common frames omitted

### Additional context

_No response_

### Environment

docker image: mcr.microsoft.com/playwright/java:v1.50.0-jammy

ArcticFoox:
@yury-s
In some environments, such as Docker or the sbt plugin, accessing the ContextClassLoader is not reliable.
Because the driver package is bundled within the JAR, class.getClassLoader() is used instead.

issues: #1447 , #1803

Fixes #1803


ArcticFoox:
Ok, I will

ArcticFoox:
Hi, @yury-s
In our internal project, we use Playwright to take a screenshot asynchronously using CompletableFuture when Open Graph tags are not found on a given page. This setup works fine locally.

However, when running the same code inside a Docker container, the following issue occurs:

```java
Thread.currentThread().getContextClassLoader() returns null
This causes failures when attempting to load resources (e.g., getResource(...)).
```
After investigation, I found that switching to:

```java
this.getClass().getClassLoader()
resolves the issue both locally and in the Docker container.
```

To help reproduce and compare the issue, I’ve created a sample repo here:
👉 https://github.com/ArcticFoox/driverJar-test

You can build the Docker image using the included Dockerfile and test the endpoints:

http://localhost:8080/playwright?url=https://google.com (uses Thread.currentThread().getContextClassLoader())

http://localhost:8080/playwright2?url=https://google.com (uses this.getClass().getClassLoader())

This is my first open source contribution, so I’d appreciate any feedback or suggestions!

Thank you 🙇

yury-s:
The test has quite some logic which make it not obvious wether the problem is in the test or in Playwright. Can you try to reduce it to a minimal example and drop all unrelated logic? Ideally it would be converted into one of the [Playwright tests](https://github.com/microsoft/playwright-java/tree/main/playwright/src/test/java/com/microsoft/playwright).

If the setup requires a custom environment (e.g. spring framework), you can put it in its own project. There is already a bunch of custom tests in the [tools/](https://github.com/microsoft/playwright-java/tree/main/tools) directory, particularly [this one](https://github.com/microsoft/playwright-java/tree/main/tools/test-spring-boot-starter). You  could convert yours into a similar one.

We also run Docker tests for our container on GitHub Actions: https://github.com/microsoft/playwright-java/actions/workflows/test_docker.yml. If the test requires Docker environment, we can run the new test inside the Playwright Docker image as part of the same test suite. Would that work?

ArcticFoox:
Sure, I’ll go ahead with that.

ArcticFoox:
Hi, @yury-s 
Not sure if this is the right approach, but I created a new test-classloader package under the tools directory for this.

The test consists of two approaches:
	1.	The first attempts to load the resource using Thread.currentThread().getContextClassLoader(),
	2.	and the second uses com.microsoft.playwright.impl.driver.jar.DriverJar.class.getClassLoader().

To verify the behavior in the Docker environment, I followed the test_docker.yml setup:
	•	Built the Docker image using utils/docker/build.sh --arm64 noble playwright-java:localbuild-noble
	•	Then ran the container via:

CONTAINER_ID="$(docker run --rm -e CI --ipc=host -v $(pwd):/root/playwright --name playwright-docker-test -d -t playwright-java:localbuild-noble /bin/bash)"

java -jar /mypath/playwright/tools/test-spring-classloader/target/test-spring-classloader-1.50.0-SNAPSHOT.jar

As a result, the first approach failed with an error, while the second one succeeded, confirming the issue.
Additionally, here’s the link to my repo with a minimal reproduction of the issue: https://github.com/ArcticFoox/driverJar-test

ArcticFoox:

tools/test-spring-classloader/src/main/java/com/microsoft/playwright/springboottest/TestApp.java
```diff
@@ -0,0 +1,92 @@
...
+
+
+    System.out.println("Starting new Playwright test...");
+
+    // Set the new driver implementation to use the DriverJar class
+    System.setProperty( "playwright.driver.impl", "com.microsoft.playwright.springboottest.DriverJar" );
+
+    CompletableFuture<Void> voidCompletableFuture2 = CompletableFuture.runAsync(() -> {
+      try (Playwright playwright = Playwright.create()) {
+        System.out.println("new Playwright test started, waiting for completion...");
```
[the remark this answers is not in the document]
ArcticFoox: Understood. I’ll proceed accordingly.

tools/test-spring-classloader/src/main/java/com/microsoft/playwright/springboottest/TestApp.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.CompletableFuture;
+
+@SpringBootApplication
+public class TestApp implements CommandLineRunner {
+
+  public static void main(String[] args) {
+    SpringApplication.run(TestApp.class, args);
+  }
+
+  public void run(String... args) {
```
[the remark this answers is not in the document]
yury-s: In that case you let's just modify existing test to either always run playwright asynchronously or, if you want to keep testing sync code path as well, do it sync/async based on a command line flag and pass the flag only when running in docker. Something like this:

```java
  public static void main(String[] args) {
    if ("--async".equals(args[0])) {
      CompletableFuture<Void> voidCompletableFuture = CompletableFuture.runAsync(() -> {
        SpringApplication.run(TestApp.class, args);
      });
      voidCompletableFuture.join();
    } else {
      SpringApplication.run(TestApp.class, args);
    }
  }
```
Would that work?

yury-s:
Unrelated, if you update on the latest `main` the client certificate tests should be green, it was an expired certificate issue.

tools/test-spring-classloader/src/main/java/com/microsoft/playwright/springboottest/TestApp.java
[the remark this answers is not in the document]
ArcticFoox: When running with --async, Spring Boot uses the main thread’s context class loader to load auto-configuration classes. However, in the asynchronous thread, this context class loader is not correctly inherited, which causes it to fail to locate the configuration files.
This results in the following error: java.lang.IllegalArgumentException: No auto configuration classes found in META-INF/spring.factories.
```bash
11:48:21.637 [ForkJoinPool.commonPool-worker-1] ERROR org.springframework.boot.SpringApplication - Application run failed
java.lang.IllegalArgumentException: No auto configuration classes found in META-INF/spring.factories. If you are using a custom packaging, make sure that file is correct.
        at org.springframework.util.Assert.notEmpty(Assert.java:470)
        at org.springframework.boot.autoconfigure.AutoConfigurationImportSelector.getCandidateConfigurations(AutoConfigurationImportSelector.java:180)
        at org.springframework.boot.autoconfigure.AutoConfigurationImportSelector.getAutoConfigurationEntry(AutoConfigurationImportSelector.java:123)
        at org.springframework.boot.autoconfigure.AutoConfigurationImportSelector$AutoConfigurationGroup.process(AutoConfigurationImportSelector.java:434)
        at org.springframework.context.annotation.ConfigurationClassParser$DeferredImportSelectorGrouping.getImports(ConfigurationClassParser.java:879)
        at org.springframework.context.annotation.ConfigurationClassParser$DeferredImportSelectorGroupingHandler.processGroupImports(ConfigurationClassParser.java:809)
        at org.springframework.context.annotation.ConfigurationClassParser$DeferredImportSelectorHandler.process(ConfigurationClassParser.java:780)
        at org.springframework.context.annotation.ConfigurationClassParser.parse(ConfigurationClassParser.java:193)
        at org.springframework.context.annotation.ConfigurationClassPostProcessor.processConfigBeanDefinitions(ConfigurationClassPostProcessor.java:330)
        at org.springframework.context.annotation.ConfigurationClassPostProcessor.postProcessBeanDefinitionRegistry(ConfigurationClassPostProcessor.java:246)
        at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanDefinitionRegistryPostProcessors(PostProcessorRegistrationDelegate.java:311)
        at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:112)
        at org.springframework.context.support.AbstractApplicationContext.invokeBeanFactoryPostProcessors(AbstractApplicationContext.java:745)
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:563)
        at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:767)
        at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:759)
        at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:426)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:326)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:1311)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:1300)
        at com.microsoft.playwright.springboottest.TestApp.lambda$main$0(TestApp.java:16)
        at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)
        at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
```
While it’s possible to manually pass the main thread’s context class loader to the async thread like below, doing so defeats the original purpose of testing the class loader behavior.
```java
public static void main(String[] args) {

    if (args.length == 0) {
      SpringApplication.run(TestApp.class, args);
    }
    else {
      if ("--async".equals(args[0])) {
        ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
        CompletableFuture<Void> voidCompletableFuture = CompletableFuture.runAsync(() -> {
          Thread.currentThread().setContextClassLoader(contextClassLoader);
          SpringApplication.run(TestApp.class, args);
        });
        voidCompletableFuture.join();
      }
    }
  }
```
For this reason, I think separating the test into a different package would make the intention clearer.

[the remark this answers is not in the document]
yury-s: Sorry, I meant to just share the implementation of the actual test method, I guess it should have been something like this:

```java
  @Override
  public void run(String... args) {
    if ("--async".equals(args[0])) {
      runAsync();
    } else {
      runSync();
    }
  }

  private void runAsync() {
    CompletableFuture<Void> voidCompletableFuture = CompletableFuture.runAsync(() -> {
      runSync();
    });
    voidCompletableFuture.join();
  }


  private void runSync() {
    try (Playwright playwright = Playwright.create()) {
      BrowserType browserType = getBrowserTypeFromEnv(playwright);
      System.out.println("Running test with " + browserType.name());
      Browser browser = browserType.launch();
      Page page = context.newPage();
      System.out.println(page.evaluate("'SUCCESS: did evaluate in page'"));
    }
  }

```

Basically the logic from the existing test is in `runSync()` and reused. I don't suggest we make any changes to the class loaders.

tools/test-spring-boot-starter/src/main/java/com/microsoft/playwright/springboottest/TestApp.java
```diff
@@ -14,6 +16,24 @@ public static void main(String[] args) {
...
+    if (args.length == 0) {
+      runSync();
+    } else {
+      if ("--async".equals(args[0])) {
+        runAsync();
+      }
+      else {
+        runSync();
+      }
+    }
```
ArcticFoox, line 28: I modified the if statement to prevent an IndexOutOfBoundsException.

ArcticFoox:
Hi, @yury-s 
I’ve applied the changes and confirmed that everything is working as expected.
Please let me know if there’s anything else I should do.

tools/test-spring-boot-starter/src/main/java/com/microsoft/playwright/springboottest/TestApp.java
```diff
@@ -14,6 +16,24 @@ public static void main(String[] args) {
 
   @Override
   public void run(String... args) {
+    if (args.length == 0) {
```
yury-s, line 19: How about the following, so that each version is called from exactly one branch?

```java
    if (Arrays.asList(args).contains("--async")) {
      runAsync();
    } else {
      runSync();
    }
```

yury-s (approved):

yury-s (approved):

.github/workflows/test_docker.yml
```diff
@@ -36,3 +36,9 @@ jobs:
         run: |
           CONTAINER_ID="$(docker run --rm -e CI --ipc=host -v $(pwd):/root/playwright --name playwright-docker-test -d -t playwright-java:localbuild-${{ matrix.flavor }} /bin/bash)"
           docker exec "${CONTAINER_ID}" /root/playwright/tools/test-local-installation/create_project_and_run_tests.sh
```
yury-s, line 38: Let's just add `docker stop "${CONTAINER_ID}"` to this (and the new) step to shutdown the container after the test has finished. Or better yet, since the container is the same in both steps, you can move its creation in a separate step and then use its CONTAINER_ID without relaunching, something like

```yaml
 - name: Start container
        run: |
          CONTAINER_ID=$(docker run --rm -e CI --ipc=host -v "$(pwd)":/root/playwright --name playwright-docker-test -d -t playwright-java:localbuild-${{ matrix.flavor }} /bin/bash)
          echo "CONTAINER_ID=$CONTAINER_ID" >> $GITHUB_ENV

      - name: Run test in container
        run: |
          docker exec "$CONTAINER_ID" /root/playwright/tools/test-local-installation/create_project_and_run_tests.sh

      - name: Test ClassLoader
        run: |
          docker exec "${CONTAINER_ID}" /root/playwright/tools/test-spring-boot-starter/package_and_run_async_test.sh
```

.github/workflows/test_docker.yml
```diff
@@ -36,3 +36,9 @@ jobs:
...
+      - name: Test ClassLoader
+        env:
+          BROWSER: ${{ matrix.browser }}
```
yury-s, line 41: nit: drop this env var, it's not used in the test

yury-s (approved):

tools/test-spring-boot-starter/src/main/java/com/microsoft/playwright/springboottest/TestApp.java
ArcticFoox: I’m glad I had the opportunity to contribute. Since this was my first open source contribution, I struggled quite a bit, but your reviews and comments were a huge help. Thank you!

.github/workflows/test_docker.yml
ArcticFoox: I added example code and stop step

.github/workflows/test_docker.yml
```diff
@@ -32,7 +32,20 @@ jobs:
...
+            CONTAINER_ID=$(docker run --rm -e CI --ipc=host -v "$(pwd)":/root/playwright --name playwright-docker-test -d -t playwright-java:localbuild-${{ matrix.flavor }} /bin/bash)
+            echo "CONTAINER_ID=$CONTAINER_ID" >> $GITHUB_ENV
+
+      - name: Run test in container
+        run: |
+            docker exec "$CONTAINER_ID" /root/playwright/tools/test-local-installation/create_project_and_run_tests.sh
+
+      - name: Test ClassLoader
+        run: |
+            docker exec "${CONTAINER_ID}" chmod +x /root/playwright/tools/test-spring-boot-starter/package_and_run_async_test.sh
```
ArcticFoox, line 46: I added a permission grant command to the workflow to resolve the permission issue.

ArcticFoox:
Hi, @yury-s
I added executable bit to package_and_run_async_test.sh using chmod +x and committed the change.

yury-s (approved):

PR #1827, opened 2025-08-18
chore: roll 1.55.0

Skn0tt:
Closes https://github.com/microsoft/playwright-java/issues/1819


playwright/src/test/java/com/microsoft/playwright/TestLaunch.java
Skn0tt, line 1: These tests were replaced upstream in https://github.com/microsoft/playwright/pull/36472. I dropped it here, don't think there's a big value from having it here.

yury-s: Let's port the new tests, at least some of them? Otherwise there is 0 coverage for that in java.

Skn0tt:
Mirroring the offline discussion here:
- the extension tests depend on having service workers available
- service workers aren't available in Java
- so we don't test it

Issue #1829, opened 2025-08-19
[internal] Use Real Trace Viewer for tracing tests


mxschmitt:
### Please do not submit this issue.

- [x] I understand

PR #1830, opened 2025-08-20
chore: migrate Trace Viewer tests to use real Trace viewer

mxschmitt:
Fixes https://github.com/microsoft/playwright-java/issues/1829

### DIFF
--- a/driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java
+++ b/driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java
@@ -40,6 +40,7 @@ public class DriverJar extends Driver {
     driverTempDir = alternativeTmpdir == null
       ? Files.createTempDirectory(prefix)
       : Files.createTempDirectory(Paths.get(alternativeTmpdir), prefix);
+    driverTempDir.toFile().deleteOnExit();
     String nodePath = System.getProperty("playwright.nodejs.path");
     if (nodePath != null) {
       preinstalledNodePath = Paths.get(nodePath);
@@ -86,11 +87,10 @@ public class DriverJar extends Driver {
     }
     ProcessBuilder pb = createProcessBuilder();
     pb.command().add("install");
-    logMessage("Executing: " + String.join(" ", pb.command()));
     pb.redirectError(ProcessBuilder.Redirect.INHERIT);
     pb.redirectOutput(ProcessBuilder.Redirect.INHERIT);
     Process p = pb.start();
-    boolean result = p.waitFor(30, TimeUnit.MINUTES);
+    boolean result = p.waitFor(10, TimeUnit.MINUTES);
     if (!result) {
       p.destroy();
       throw new RuntimeException("Timed out waiting for browsers to install");
--- a/playwright/src/test/java/com/microsoft/playwright/TraceViewerPage.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TraceViewerPage.java
@@ -23,42 +23,42 @@ import java.nio.file.Path;
 import com.microsoft.playwright.impl.driver.Driver;
 import com.microsoft.playwright.options.AriaRole;
 
-public class TraceViewerPage {
+class TraceViewerPage {
   private final Page page;
 
-  public TraceViewerPage(Page page) {
+  TraceViewerPage(Page page) {
     this.page = page;
   }
 
-  public Page page() {
+  Page page() {
     return page;
   }
 
-  public Locator actionsTree() {
+  Locator actionsTree() {
     return page.getByTestId("actions-tree");
   }
 
-  public Locator actionTitles() {
+  Locator actionTitles() {
     return page.locator(".action-title");
   }
 
-  public Locator stackFrames() {
+  Locator stackFrames() {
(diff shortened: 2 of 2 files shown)


playwright/src/test/java/com/microsoft/playwright/Server.java
```diff
@@ -187,7 +193,19 @@ public void handle(HttpExchange exchange) throws IOException {
       path = "/index.html";
     }
 
-    // Resources from "src/test/resources/" are copied to "resources/" directory in the jar.
+    // If static files directory is set, serve from filesystem first
+    if (staticFilesDirectory != null) {
```
mxschmitt, line 197: underneath we do something similar but via `getResourceAsStream` which got introduced [here](https://github.com/microsoft/playwright-java/pull/657/files#diff-e20e283060b38eabf5fb0339495c168fa6545597a11be0aa13adc3760433514aR196) - do you remember why?

yury-s: This is the standard way to read resources in a maven package (they move from the source into the test package artifact). I'd keep using it with the custom base dir too, can we do that?

playwright/src/test/java/com/microsoft/playwright/Server.java
```diff
@@ -187,7 +193,19 @@ public void handle(HttpExchange exchange) throws IOException {
...
+      java.nio.file.Path filePath = staticFilesDirectory.resolve(path.substring(1)); // Remove leading /
+      if (java.nio.file.Files.exists(filePath) && !java.nio.file.Files.isDirectory(filePath)) {
```
yury-s, line 199: Drop `java.nio.file.` prefix?

playwright/src/test/java/com/microsoft/playwright/TestTracing.java
```diff
@@ -284,41 +357,95 @@ public void shouldNotRecordNetworkActions(@TempDir Path tempDir) throws IOExcept
...
-    assertEquals(asList("Frame.goto"), calls);
+    showTraceViewer(traceFile1, traceViewer -> {
+      assertThat(traceViewer.actionTitles()).hasText(new Pattern[] {
+        Pattern.compile("Navigate to \"/empty.html\"")
+      });
+    });
+  }
+
+  private void showTraceViewer(Path tracePath, TraceViewerConsumer callback) throws Exception {
+    Path driverDir = Driver.ensureDriverInstalled(java.util.Collections.emptyMap(), true).driverDir();
```
yury-s, line 368: Move this method along with the trace viewer POM into its own file? Also consider turning the viewer into a fixture.

playwright/src/test/java/com/microsoft/playwright/Server.java
mxschmitt: I think there is no way of accessing the trace-viewer html source code via the tests/resource directory so I introduced another resourceLoader concept - are you fine with that?

yury-s: Yes. I'd pass the resource provide in server the constructor.

driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java
```diff
@@ -87,10 +86,11 @@ private void installBrowsers(Map<String, String> env) throws IOException, Interr
     }
     ProcessBuilder pb = createProcessBuilder();
     pb.command().add("install");
+    logMessage("Executing: " + String.join(" ", pb.command()));
     pb.redirectError(ProcessBuilder.Redirect.INHERIT);
     pb.redirectOutput(ProcessBuilder.Redirect.INHERIT);
     Process p = pb.start();
-    boolean result = p.waitFor(10, TimeUnit.MINUTES);
+    boolean result = p.waitFor(30, TimeUnit.MINUTES);
```
yury-s, line 93: why is 10 not enough?

driver-bundle/src/main/java/com/microsoft/playwright/impl/driver/jar/DriverJar.java
```diff
@@ -40,7 +40,6 @@ public DriverJar() throws IOException {
     driverTempDir = alternativeTmpdir == null
       ? Files.createTempDirectory(prefix)
       : Files.createTempDirectory(Paths.get(alternativeTmpdir), prefix);
-    driverTempDir.toFile().deleteOnExit();
```
yury-s, line 43: Revert? Or who's going to delete the mess?

playwright/src/test/java/com/microsoft/playwright/TraceViewerFixture.java
yury-s, line 1: Is it some failed attempt? If so, let's delete the empty file.

playwright/src/test/java/com/microsoft/playwright/TraceViewerPage.java
```diff
@@ -0,0 +1,119 @@
...
+package com.microsoft.playwright;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import com.microsoft.playwright.impl.driver.Driver;
+import com.microsoft.playwright.options.AriaRole;
+
+public class TraceViewerPage {
```
yury-s, line 26: nit: you can drop all `public` specifiers.

PR #1835, opened 2025-08-26
chore: rename headful -> headed

yury-s:
This is how we call it everywhere else


playwright/src/test/java/com/microsoft/playwright/TestBase.java
```diff
@@ -59,7 +59,7 @@ public class TestBase {
   BrowserContext context;
 
   static boolean isHeadful() {
-    return headful;
+    return headed;
```
mxschmitt, line 62: nit: rename method to isHeaded?

mxschmitt (approved):

PR #1886, opened 2026-01-27
fix: handle null close code and reason in WebSocketRoute

### LINKED ISSUE #1882: [Bug]: NullPointerException in WebSocketRouteImpl when page redirects during WebSocket message handling
### Version

1.57.0

### Steps to reproduce

TODO a test repo

Steps to reproduce:

1. Set up a WebSocket route handler using page.routeWebSocket()
2. Have the WebSocket receive a message that triggers a page navigation/redirect
3. The redirect closes the page while the WebSocket handler is still processing

### Expected behavior

The WebSocket connection should close gracefully when the page redirects, without throwing exceptions.

### Actual behavior

Actual behavior:
A `NullPointerException` is thrown in `WebSocketRouteImpl.java` at `int code = params.get("code").getAsInt();` because code and reason are null when the connection is closed due to page navigation.

### Additional context

**Analysis:**
The JavaScript implementation doesn't send this data:
https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/dispatchers/webSocketRouteDispatcher.ts#L93

The Java implementation doesn't have null guards:
https://github.com/microsoft/playwright-java/blob/main/playwright/src/main/java/com/microsoft/playwright/impl/WebSocketRouteImpl.java#L163-L165

**Suggested fix:**
Add null checks or default values for `code` and `reason` in the close handler.

### Environment

- We use the docker image 
- OS: Linux
- Java 21
- All browsers

nanne-rl:
When a WebSocket connection closes due to page navigation, the driver may send close events without code, reason, or wasClean fields. This caused a NullPointerException when accessing these fields directly.

Fixes #1882

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/WebSocketRouteImpl.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/WebSocketRouteImpl.java
@@ -160,9 +160,9 @@ class WebSocketRouteImpl extends ChannelOwner implements WebSocketRoute {
         sendMessageAsync("sendToPage", messageParams);
       }
     } else if ("closePage".equals(event)) {
-      int code = params.get("code").getAsInt();
-      String reason = params.get("reason").getAsString();
-      boolean wasClean = params.get("wasClean").getAsBoolean();
+      Integer code = params.has("code") ? params.get("code").getAsInt() : null;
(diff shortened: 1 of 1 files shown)


yury-s (approved):
Consider adding a test to `TestRouteWebSocket`

PR #1920, opened 2026-05-18
Add code coverage API

DF-Nico:
Using CDP Session, provide simple API to extract code coverage from JS/CSS

Still in WIP in some point, but before going further, if you already have some feedback, I'm all hears


DF-Nico:

yury-s:
Start with filing an issue, I don't think we want to port this API from node.

PR #1927, opened 2026-06-12
feat: assemble driver from npm instead of CDN

yury-s:
## What

Download the driver from the canonical npm registry rather than `cdn.playwright.dev`. Each platform's driver is now **assembled** from the [`playwright-core`](https://www.npmjs.com/package/playwright-core) npm package plus the matching Node.js binary from nodejs.org, the same way the upstream Playwright build does it — mirroring the approach in [microsoft/playwright-dotnet#3322](https://github.com/microsoft/playwright-dotnet/pull/3322).

The driver version is pinned to the exact upstream commit via `npm view playwright@<version> gitHead`, and the paired `NODE_VERSION` is read from that commit's `build-playwright-driver.sh`.

## api.json

The npm package does not contain `api.json` (it is a build artifact upstream copies into the CDN zip). So `generate_api.sh` now generates it from the upstream Playwright source at the same commit — via a minimal sparse checkout (`utils` + `docs`, no `npm install`) running `generateApiJson.js` — instead of relying on the bundled driver's `print-api-json`. Set `PW_SRC_DIR` to reuse an existing `microsoft/playwright` checkout.

## Verification

- All 5 platforms (`mac`, `mac-arm64`, `linux`, `linux-arm64`, `win32_x64`) assemble in ~12s with correct per-arch Node.js binaries.
- The assembled `package/` is **byte-identical** to the official CDN zip's, except the two intentionally-dropped build artifacts (`api.json`, `protocol.yml` — the latter unused at runtime).
- The driver runs via the real runtime path (`node package/cli.js --version`).
- The generated `api.json` is **byte-identical** to the official CDN-bundled 1.60.0 `api.json`.

References https://github.com/microsoft/playwright-internal/issues/291

### DIFF
--- a/scripts/download_driver.sh
+++ b/scripts/download_driver.sh
@@ -8,8 +8,10 @@ cd "$(dirname $0)"
 
 if [[ ($1 == '-h') || ($1 == '--help') ]]; then
   echo ""
-  echo "This script for downloading playwright driver for all platforms."
-  echo "The downloaded files will be put under 'driver-bundle/src/main/resources/driver'."
+  echo "This script downloads and assembles the Playwright driver for all platforms."
+  echo "Each driver is assembled from the 'playwright-core' npm package and the matching"
+  echo "Node.js binary from https://nodejs.org, the same way the upstream Playwright build"
+  echo "does it. The result is put under 'driver-bundle/src/main/resources/driver'."
   echo ""
   echo "Usage: scripts/download_driver.sh [option]"
   echo ""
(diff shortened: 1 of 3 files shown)


scripts/download_driver.sh
```diff
@@ -19,8 +21,40 @@ if [[ ($1 == '-h') || ($1 == '--help') ]]; then
...
+# Resolve the exact upstream commit that produced this driver version, so that the
+# bundled Node.js version matches the driver exactly.
+GIT_HEAD=$(npm view playwright@"$DRIVER_VERSION" gitHead)
+if [[ -z "$GIT_HEAD" ]]; then
+  echo "Failed to resolve upstream commit (gitHead) for playwright@$DRIVER_VERSION"
+  exit 1
+fi
+
+# The Node.js version is kept in sync with the driver version in the upstream build script.
+NODE_VERSION=$(curl -fsSL "https://raw.githubusercontent.com/microsoft/playwright/$GIT_HEAD/utils/build/build-playwright-driver.sh" \
```
dgozman, line 48: I think we'll get rid of this script later on, so the node version would be somewhere else (probably hardcoded and updated right here in this script).

dgozman (approved):

scripts/download_driver.sh
yury-s: I'm on the fence, we have this auto-updater upstream which could jus bump NODE_VERSION file there. In any case we can update this line later on.

PR #1946, opened 2026-07-05
feat(driver) Introduce experimental support for third-party Playwright drivers

antohaby:
Hey, Kotlin Gradle Plugin (KGP) maintainer here.

# Problem

We at KGP working on integration of Playwright as tool to control browser and verify kotlin code in them.
Gradle works on JVM so we found Playwright-java extremely useful as it allow controlling browsers natively in JVM.
KGP already integrated with Node JS and NPM tooling, but Playwright-java comes with its own node.js instances and js-written driver in a driver bundle jar. The abstract Driver offers some customisation via System properties and env variables.
But it doesn't really decouples Playwright-client with the Driver.
It also bound instance to static field, making it immutable within class loader.

We would like to run Playwright with custom managed driver instance.

In the simplest case we would like to create Driver instance explicitly, without polluting System properties or relying on env variables. Customize creation of the ProcessBuilder by providing paths to our Node.JS and playwright-core NPM package  locations.

# Solution

Introduce Driver.ThirdPartyDriver marker-interface and create separate factory method to be called with instance of the driver marked with this interface.

This provides strong separation between default drivers and the ones that passed by user. These drivers can be specially handled when needed. Since, by definition, these drivers may not work as expected.

Add two interfaces of driver variants:
1. External Process builder.
2. And more generic ByteChannel driver.

Add two smoke tests to verify ability of PW to communicate with custom drivers.

# Options
I understand that creation of 3 new interfaces may not be so necessary for our case, Instead I can offer a simpler change, where Playwright.createImpl() can accept just instance of "abstract class Driver" with custom process builder.

Although I think it makes sense to have better control over for 3rd party drivers, if they will evolve. Marker interface can be useful.

### DIFF
--- a/driver/src/main/java/com/microsoft/playwright/impl/driver/Driver.java
+++ b/driver/src/main/java/com/microsoft/playwright/impl/driver/Driver.java
@@ -16,6 +16,7 @@
 
 package com.microsoft.playwright.impl.driver;
 
+import java.nio.channels.ByteChannel;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


antohaby:
@microsoft-github-policy-service agree [company="JetBrains"]

antohaby:
@microsoft-github-policy-service agree company="JetBrains"

yury-s:
Can you start with opening an issue that describes your scenario and the requirements? The driver is an implementation detail of playwright and in general we want it to be bundled with the java part as the driver's API is not compatible across versions.

yury-s:
This might be a valid use case that we'd want to address. The current description in terms of Playwright implementation details makes it hard to understand what the ultimate goal is and the change doesn't seem to be at the right place. Please open an issue that describes your use case, current constraints and what is lacking in the current **public** API of Playwright.

PR #1947, opened 2026-07-05
feat: add Automatic-Module-Name manifest entries

raphw:
Give the published jars stable Java module names so they can be required from the module path instead of relying on filename-derived automatic names:

- playwright: com.microsoft.playwright
- driver: com.microsoft.playwright.driver
- driver-bundle: com.microsoft.playwright.driver.bundle

The playwright test-jar carries its own name
(com.microsoft.playwright.test) so the main jar and the test-jar never claim the same module.

### DIFF
--- a/driver-bundle/pom.xml
+++ b/driver-bundle/pom.xml
@@ -16,6 +16,10 @@
     Node.js is preinstalled on the host (see PLAYWRIGHT_NODEJS_PATH).
   </description>
 
+  <properties>
+    <automatic.module.name>com.microsoft.playwright.driver.bundle</automatic.module.name>
+  </properties>
+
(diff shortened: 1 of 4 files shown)


raphw:

yury-s:
What practical issue does it solve?

yury-s (approved):

raphw:
It avoids that Playwright depends on file names when you use Java's module system and want to use Playwright from this application. 

Right now two modular applications might choose different names and become incompatible if they choose different file names and therefore module names. If Playwright makes this choice it provides a stable name for consumption.

Also, some tooling requires a module name at least on the manifest level.

PR #1948, opened 2026-07-09
fix(screencast): register AnnotatePosition enum serializer

### LINKED ISSUE #1912: [Bug]: showActions() with ShowActionsOptions.setPosition() throws protocol error for all AnnotatePosition values
### Version

1.59.0

### Steps to reproduce

page.screencast().showActions(                                                                                                                                                                                                                                            
      new Screencast.ShowActionsOptions().setPosition(AnnotatePosition.TOP_RIGHT)
  );


### Expected behavior

  showActions() runs without error and action indicators are displayed in the screencast at the specified position.  

### Actual behavior

 Actual Behavior                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                            
  A protocol-level error is thrown for at least TOP and TOP_RIGHT - i haven't tried every single one.

 Error                                                                                                                                                                                                                                                                       
   
  Error: position: expected one of (top-left|top|top-right|bottom-left|bottom|bottom-right)                                                                                                                                                                                   
    name='Error                                                                                                                                                                                                                                                             
    stack='Error: position: expected one of (top-left|top|top-right|bottom-left|bottom|bottom-right)
      at .../playwright-java-.../package/lib/protocol/validatorPrimitives.js:156:13                                                                                                                                                                                           
      at .../playwright-java-.../package/lib/protocol/validatorPrimitives.js:122:12
      at .../playwright-java-.../package/lib/protocol/validatorPrimitives.js:140:21                                                                                                                                                                                           
      at DispatcherConnection.dispatch (.../dispatcher.js:264:21)   

### Additional context
Directly using a String as per the Documentation, like
      new Screencast.ShowActionsOptions().setPosition("top-right")
does not work, as the Method Signature doesn't even allow it.

### Environment

 - playwright-java version: 1.59.0
  - OS: macOS 14
  - Java version: 17

Develop-KIM:
## Summary
- `AnnotatePosition` had no Gson serializer registered, so `page.screencast().showActions(new Screencast.ShowActionsOptions().setPosition(AnnotatePosition.TOP_RIGHT))` sent the raw constant `TOP_RIGHT` to the driver, which rejected it: `position: expected one of (top-left|top|top-right|...)`.
- Registered it with `ToLowerCaseAndDashSerializer` (as used for `ColorScheme`, `ServiceWorkerPolicy`, …) so values serialize as `top-right` etc.

Fixes https://github.com/microsoft/playwright-java/issues/1912


Develop-KIM:

playwright/src/test/java/com/microsoft/playwright/impl/TestSerialization.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.microsoft.playwright.options.AnnotatePosition;
+import org.junit.jupiter.api.Test;
+
+import static com.microsoft.playwright.impl.Serialization.gson;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestSerialization {
+  @Test
+  void annotatePositionSerializesToLowerCaseAndDash() {
```
yury-s: let's turn it into and e2e test and check that actual playwright api call doesn't throw `page.screencast().showActions(new Screencast.ShowActionsOptions().setPosition(AnnotatePosition.TOP_RIGHT))`

yury-s (approved):
The change looks good, let's update the test

Develop-KIM:
Done — dropped the serializer unit test and moved the coverage into `TestScreencast`, going through the real API:

```java
for (AnnotatePosition position : AnnotatePosition.values()) {
  AutoCloseable disposable = page.screencast().showActions(
    new Screencast.ShowActionsOptions().setPosition(position));
  assertNotNull(disposable);
  disposable.close();
}
```

Ran it against a real browser both ways. Without the serializer registration every position fails with `position: expected one of (top-left|top|top-right|bottom-left|bottom|bottom-right)`, which is the error from #1912; with it the test passes.

I looped over all six values rather than just `TOP_RIGHT` so a future enum constant that doesn't round-trip gets caught too — happy to narrow it if you'd rather keep the test minimal.

yury-s (approved):

PR #1954, opened 2026-07-29
chore(driver): roll to 1.62.0

### LINKED ISSUE #38919: [Bug]: locale is not propagated to workers in firefox
### Version

95a728a037070f31afb7decebbc4d9e2284e26ba

### Steps to reproduce

https://github.com/microsoft/playwright/blob/95a728a037070f31afb7decebbc4d9e2284e26ba/tests/library/browsercontext-locale.spec.ts#L166

This test passes in CI, but not locally. I assume it's because CI locale happens to give the same output as es-MX. If you change the locale to ru-RU, it should also fail in CI.

It seems that workers are not respecting the context locale.

### Expected behavior

Workers should be respecting context locale.

### Actual behavior

Workers are using system locale.

### Additional context

found during https://github.com/microsoft/playwright-python/pull/3026

### Environment

```shell
System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 133.67 MB / 18.00 GB
  Binaries:
    Node: 20.19.0 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/node
    Yarn: 1.22.22 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/yarn
    npm: 9.9.4 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/npm
    pnpm: 10.28.1 - /Users/skn0tt/.local/state/fnm_multishells/66603_1769161758461/bin/pnpm
    Deno: 2.5.2 - /opt/homebrew/bin/deno
  IDEs:
    VSCode: 1.108.1 - /opt/homebrew/bin/code
    Claude Code: 2.1.15 - /opt/homebrew/bin/claude
  Languages:
    Bash: 3.2.57 - /bin/bash
```

dcrousso:
Rolls the driver to `1.62.0` and regenerates the Java interfaces from the `v1.62.0` documentation.

Upstream removed the `routeFromHAR` `interceptAPIRequests` option before `1.62.0` in <https://github.com/microsoft/playwright/pull/41964> and <https://github.com/microsoft/playwright/pull/41965>.  This drops the generated `interceptAPIRequests` field, the `setInterceptAPIRequests` builder, the `HARRouter.addAPIRequestRoute` helper, the `routeAPIRequestsFromHar` and `unrouteAPIRequestsFromHar` calls, and their test.  `shouldNotInterceptAPIRequestContextRequestsByDefault` remains because that behavior still holds.

Firefox `153` fixes <https://github.com/microsoft/playwright/issues/38919>, so the worker locale test now uses the context locale expectation for every browser.

Validation:

- `mvn compile` succeeds
- `mvn test -f playwright` runs `1696` tests with `26` skipped tests
- `TestBrowserContextHar` passes with `27` tests
- `TestWorkers` passes on Firefox and Chromium
- API generation used `PW_SRC_DIR` pinned to `v1.62.0` at `e3950d9c140d007bd52853b45813c6274b24e36f`

`TestChromiumTracing.shouldSupportABufferWithoutAPath` reproduces `PASS FAIL PASS` on both this branch and the prior driver, while passing consistently in isolation.


scripts/DRIVER_VERSION
```diff
@@ -1 +1 @@
-1.62.0-alpha-2026-07-22
+1.62.0
```
yury-s, line 1: we should probably roll 1.62.1 before the release

yury-s (approved):

PR #1958, opened 2026-08-03
chore(release): mark 1.62.0

dcrousso:
The Maven projects remained at `1.50.0-SNAPSHOT` after the driver roll.

Sets every project version and the example dependency to `1.62.0` using `./scripts/set_maven_version.sh 1.62.0`.


yury-s (approved):
Let's update driver to 1.62.1 first?

PR #1965, opened 2026-08-17
fix(evaluate): 0.0 is serialized as -0

ARDA7787:


yury-s (commented):
Can you add a test for this?

yury-s:
Thanks for spotting this! Superseded by #1981, which includes the same fix along with a test.

PR #1967, opened 2026-08-25
fix(tracing): send addStackToTracingNoReply on the LocalUtils connection

### LINKED ISSUE #1962: [Bug]: Tracing setSources(true) produces no sources when the browser is connected via connect()
### Version

1.61.0

### Steps to reproduce

1. `src/main/java/org/example/TraceSources.java`:

```java
package org.example;

import com.microsoft.playwright.Browser;
import com.microsoft.playwright.BrowserContext;
import com.microsoft.playwright.Page;
import com.microsoft.playwright.Playwright;
import com.microsoft.playwright.Tracing;
import java.nio.file.Paths;

public class TraceSources {
  public static void main(String[] args) {
    String wsEndpoint = args.length > 0 ? args[0] : null;

    try (Playwright playwright = Playwright.create()) {
      Browser browser = wsEndpoint == null
          ? playwright.chromium().launch()
          : playwright.chromium().connect(wsEndpoint);

      context.tracing().start(new Tracing.StartOptions()
          .setScreenshots(true)
          .setSnapshots(true)
          .setSources(true));

      Page page = context.newPage();
      page.navigate("https://playwright.dev");
      page.getByText("Get started").click();

      context.tracing().stop(new Tracing.StopOptions()
          .setPath(Paths.get(wsEndpoint == null ? "trace-launch.zip" : "trace-connect.zip")));
      browser.close();
    }
  }
}
```

2. Point PLAYWRIGHT_JAVA_SRC at the source root, then run once with a locally
   launched browser and once against a Playwright server, e.g. if Maven project:

```
export PLAYWRIGHT_JAVA_SRC=src/main/java

# A: locally launched browser
mvn -q compile exec:java -Dexec.mainClass=org.example.TraceSources

# B: connected browser. Start the server in another shell first:
#   npx playwright@1.61.0 run-server --port 3000
mvn -q compile exec:java -Dexec.mainClass=org.example.TraceSources \
    -Dexec.args="ws://localhost:3000/"
```

4. Open both in the trace viewer:

```
mvn exec:java -Dexec.mainClass=com.microsoft.playwright.CLI -Dexec.args="show-trace trace-launch.zip"

mvn exec:java -Dexec.mainClass=com.microsoft.playwright.CLI -Dexec.args="show-trace trace-connect.zip"
```

### Expected behavior

The Source tab in the trace viewer shows TraceSources.java with the selected action's line highlighted, for both traces.

### Actual behavior

A (launch)
Source tab: shows TraceSources.java, current action highlighted.

B (connect)
Source tab: empty. Actions still show a file name and line number, but the file
content is not in the archive, so no code is displayed.

No exception is thrown and nothing is logged. The trace is written successfully and
is otherwise complete — screenshots, snapshots and network are all there. Only the
sources are missing.

### Additional context

Possible cause:
We have not confirmed this, but source collection looks like it happens purely
client-side: the per-call Java stacks are pushed to the local driver's LocalUtils via
`addStackToTracingNoReply`, and `stop()` writes `trace.stacks` and the
`resources/src@*.txt` blobs from that stack session. `Connection.internalSendMessage`
appears to send that message over the same transport as the API call it belongs to,
which for a connected browser is the remote connection rather than the local driver.
If so, the local stack session stays empty and both entries are skipped.

### Environment

- Operating System: Ubuntu 22.04
- CPU: x64
- Browser: Chromium (maybe the others as well)
- Java Version: 21

Develop-KIM:
## Summary

- `Connection.internalSendMessage` addressed `addStackToTracingNoReply` to `localUtils.guid` but sent it over its own transport. For a browser from `connect()` that transport is the server pipe, not the driver connection that owns `LocalUtils`, so the stack session opened by `tracingStarted` never received anything and `setSources(true)` produced no sources.
- Sends it on `localUtils.connection` instead, matching `this._localUtils?.addStackToTracingNoReply(...)` in `packages/playwright-core/src/client/connection.ts`. For a local connection that is the same object, so nothing changes there.
- `TestBrowserTypeConnect#shouldRecordTraceWithSources` already covered this and fails on `main` (0 sources instead of 1); it passes with the change. It never ran on CI because the sources job filters on `-D test=*TestTracing*`.

Fixes https://github.com/microsoft/playwright-java/issues/1962

### DIFF
--- a/playwright/src/main/java/com/microsoft/playwright/impl/Connection.java
+++ b/playwright/src/main/java/com/microsoft/playwright/impl/Connection.java
@@ -190,7 +190,8 @@ public class Connection {
       callData.add("stack", stack);
       JsonObject stackParams = new JsonObject();
       stackParams.add("callData", callData);
-      internalSendMessage(localUtils.guid,"addStackToTracingNoReply", stackParams, null, false, true);
+      // LocalUtils belongs to the driver connection, which is not this one when the browser came from connect().
+      localUtils.connection.internalSendMessage(localUtils.guid,"addStackToTracingNoReply", stackParams, null, false, true);
     }
     return result;
   }
--- a/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestBrowserTypeConnect.java
@@ -45,7 +45,7 @@ public class TestBrowserTypeConnect extends TestBase {
   private Process browserServer;
   private String wsEndpoint;
 
-  private static class BrowserServer {
+  static class BrowserServer {
     Process process;
     String wsEndpoint;
 
@@ -61,7 +61,7 @@ public class TestBrowserTypeConnect extends TestBase {
     }
   }
 
-  private static BrowserServer launchBrowserServer(BrowserType browserType) {
+  static BrowserServer launchBrowserServer(BrowserType browserType) {
     try {
       Driver driver = Driver.ensureDriverInstalled(Collections.emptyMap(), false);
       Path dir = driver.driverDir();
--- a/playwright/src/test/java/com/microsoft/playwright/TestTracingOverConnect.java
+++ b/playwright/src/test/java/com/microsoft/playwright/TestTracingOverConnect.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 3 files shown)


yury-s (approved):
Do you mind adding a test?

Develop-KIM:
Added a test: `TestTracingOverConnect` connects to a launch-server browser and traces with `setSources(true)`. It passes the source root through `Playwright.CreateOptions` instead of reading `PLAYWRIGHT_JAVA_SRC` from the environment, so it runs in the regular test job, not only the `*TestTracing*` sources one. Without the fix it gets 0 sources instead of 1.

yury-s (approved):

Develop-KIM:
The CI failure came from main rolling the driver to 1.63 (#1972) after this branch was cut: the trace zip now stores sources as `src/<sha1>.java` instead of `resources/src@<sha1>.txt`, so the `.txt` filter in the new test found nothing. Switched it to the `startsWith("src/")` check that `TestJavaSourceLocationInConstructor` got in that roll, and gave `shouldRecordTraceWithSources` in `TestBrowserTypeConnect` the same update since it had the same filter.

PR #1970, opened 2026-09-03
feat(api): add JSpecify nullability annotations

### LINKED ISSUE #473: [Feature] Add nullable reference annotations.
The playwright library has no clear documentation about whether an argument or return type can be null. In the original TypeScript library it's not a problem since the type system of TypeScript supports nullable annotation. For example, the `page.contentText` has a `string | null` type, which indicates that it may be null. With `strictNullCheck` compiler option, you can't call its methods without proper null check.

image

However, in this project, the nullable annotation is absent, neither the documentation points out the fact that it can be null.

image

Java actually has a plenty of null annotations. [These annotations can also be identify in Kotlin](https://kotlinlang.org/docs/java-interop.html#nullability-annotations). I hope nullable can be properly annotated like the TypeScript library, or at least be documented.

yury-s:
## Summary
- generate JSpecify nullability annotations for the public API
- mark public API packages as null-marked
- verify annotations with a Java NullAway consumer check

Fixes https://github.com/microsoft/playwright-java/issues/473

### DIFF
--- a/playwright/pom.xml
+++ b/playwright/pom.xml
@@ -66,6 +66,10 @@
       <artifactId>gson</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.jspecify</groupId>
+      <artifactId>jspecify</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.java-websocket</groupId>
       <artifactId>Java-WebSocket</artifactId>
     </dependency>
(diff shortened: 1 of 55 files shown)


playwright/pom.xml
```diff
@@ -99,4 +103,64 @@
       <artifactId>driver-bundle</artifactId>
     </dependency>
   </dependencies>
+  <profiles>
```
dgozman, line 106: I assume this is a test-only null check setup? Otherwise, seems very involved for production pom.

dgozman (approved):

playwright/pom.xml
yury-s: Yes, this is for tests only:

```xml
<phase>test-compile</phase>
<goals>
  <goal>testCompile</goal>
 </goals>
```

PR #1986, opened 2026-09-15
devops: move docker publishing to Azure DevOps

yury-s:
## Summary
- Replace the GitHub Actions Docker publish workflow (failed at `azure/login` for v1.63.0) with `.azure-pipelines/publish-docker.yml`, ported from upstream (microsoft/playwright#41724).
- Docker builds pull base images via the ACR cache, use the Azure Ubuntu mirror, and resolve Maven through DevDiv_PublicPackages (settings passed as a build secret); `.dockerignore` keeps the pipeline's `.npmrc` out of images.
- Add a `canary` release channel for manual pipeline runs.


.dockerignore
```diff
@@ -0,0 +1,5 @@
+# The repository root is the Docker build context (see utils/docker/build.sh).
+# Keep registry credentials written by the release pipeline out of the images.
+**/.npmrc
```
dgozman, line 3: should this include maven secrets as well?

dgozman (approved):

.dockerignore
yury-s: No need — MavenAuthenticate injects credentials into `~/.m2/settings.xml` on the agent, which is outside the build context (repo root), so `COPY .` can't pick it up. It reaches the build only via `--secret` and isn't persisted in any layer. `.npmrc` needs the entry only because the pipeline writes it into the repo root.

PR #1988, opened 2026-09-16
fix(junit): create a new Playwright when the previous launcher run closed it

### LINKED ISSUE #1751: [Bug]: Playwight-Junit lifecycle incompatible with maven retries 'surefire.rerunFailingTestsCount'
### Version

1.50.0

### Steps to reproduce

1. Have a basic Playwright test that fails
2. Try running it with maven retries: `./mvnw test -Dtest=MyTest -Dsurefire.rerunFailingTestsCount=2`

### Expected behavior

System is able to retry the failing test, system can work with Playwright context (browser, page etc.)

### Actual behavior

First attempt is a legit failure, subsequent rerun attempts have broken Playwright context
```[INFO] Running fm.seller.onboardingweb.test.MyTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.002 s <<< FAILURE! -- in fm.seller.onboardingweb.test.MyTest
[ERROR] fm.seller.onboardingweb.test.MyTest.test(Page) -- Time elapsed: 0.001 s <<< ERROR!
org.junit.jupiter.api.extension.ParameterResolutionException: Failed to resolve parameter [com.microsoft.playwright.Page page] in method [public void fm.seller.onboardingweb.test.MyTest.test(com.microsoft.playwright.Page)]: Playwright connection closed
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: com.microsoft.playwright.PlaywrightException: Playwright connection closed
        at com.microsoft.playwright.impl.PipeTransport.send(PipeTransport.java:50)
        at com.microsoft.playwright.impl.Connection.internalSendMessage(Connection.java:168)
        at com.microsoft.playwright.impl.Connection.sendMessageAsync(Connection.java:134)
        at com.microsoft.playwright.impl.ChannelOwner.sendMessageAsync(ChannelOwner.java:109)
        at com.microsoft.playwright.impl.SelectorsImpl.setTestIdAttributeName(SelectorsImpl.java:42)
        at com.microsoft.playwright.impl.SharedSelectors.lambda$setTestIdAttribute$2(SharedSelectors.java:72)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
        at com.microsoft.playwright.impl.SharedSelectors.setTestIdAttribute(SharedSelectors.java:72)
        at com.microsoft.playwright.impl.junit.ExtensionUtils.setTestIdAttribute(ExtensionUtils.java:53)
        at com.microsoft.playwright.impl.junit.BrowserContextExtension.getOrCreateBrowserContext(BrowserContextExtension.java:60)
        at com.microsoft.playwright.impl.junit.PageExtension.getOrCreatePage(PageExtension.java:51)
        at com.microsoft.playwright.impl.junit.PageExtension.resolveParameter(PageExtension.java:35)
        ... 2 more
```

```
[ERROR] Errors: 
[ERROR] fm.seller.onboardingweb.test.MyTest.test(Page)
[ERROR]   Run 1: MyTest.test:13
[ERROR]   Run 2: MyTest.test(Page) » ParameterResolution Failed to resolve parameter [com.microsoft.playwright.Page page] in method [public void fm.seller.onboardingweb.test.MyTest.test(com.microsoft.playwright.Page)]: Playwright connection closed
[ERROR]   Run 3: MyTest.test(Page) » ParameterResolution Failed to resolve parameter [com.microsoft.playwright.Page page] in method [public void fm.seller.onboardingweb.test.MyTest.test(com.microsoft.playwright.Page)]: Playwright connection closed
```

### Additional context

Minimal test:
```
@UsePlaywright(OptionsFactory.class)
public class MyTest {
    @Test
    public void test(Page page) {
        page.navigate("https://google.com/");
        assert false;
    }
}
```

### Environment

- OS name: "mac os x", version: "15.3", arch: "aarch64", family: "mac"
- Java version: 17.0.2, vendor: Oracle Corporation
- Apache Maven 3.8.5 (3599d3414f046de2324203b78ddcf9b5e4388aa0)
- playwright 1.50.0
- Junit 5.11.4
- surefire 3.5.2

Develop-KIM:
## Summary

- Surefire's `rerunFailingTestsCount` (this repo's own CI uses it through `PW_MAX_RETRIES`) runs the launcher again in the same JVM and thread. The first run's `PlaywrightRegistry` has already closed every `Playwright` it created, but the `ThreadLocal` in `PlaywrightExtension` still hands one of them out, so the rerun launches a browser on a closed connection and fails with `Playwright connection closed`.
- Reuses the `ThreadLocal` instance only when it belongs to the current run's registry and creates a fresh one otherwise. Clearing the `ThreadLocal` from `close()` would not reach other threads.
- This only shows up with JUnit 5.13+: since #1860 the registry is `AutoCloseable`, which older JUnit versions never close, so there the first `Playwright` just leaks and the rerun reuses it.
- `TestFixturesRerun` runs `RerunFixture` through the launcher twice on one thread and fails on `main` with the error above. That needs `junit-platform-launcher`, which comes in through the `junit-bom` so it follows `junit.version`.

Fixes https://github.com/microsoft/playwright-java/issues/1751


Develop-KIM:
The Docker jobs were failing because `tools/test-local-installation` compiles a copy of `playwright/src/test` with its own pom, which didn't have the `junit-platform-launcher` dependency the new test needs. Added it there in 3d8efa0; the smoke run in that project is green for me locally.

playwright/src/main/java/com/microsoft/playwright/impl/junit/PlaywrightExtension.java
```diff
@@ -85,13 +88,14 @@ public Object resolveParameter(ParameterContext parameterContext, ExtensionConte
    * @return The Playwright that belongs to the current test.
    */
   public static Playwright getOrCreatePlaywright(ExtensionContext extensionContext) {
+    PlaywrightRegistry registry = PlaywrightRegistry.getOrCreateFor(extensionContext);
     Playwright playwright = threadLocalPlaywright.get();
-    if (playwright != null) {
+    // A previous launcher run on this thread (e.g. a surefire rerun) has already closed its Playwright.
+    if (playwright != null && registry.owns(playwright)) {
```
yury-s, line 94: `PlaywrightRegistry` implements only `AutoCloseable` (it stopped implementing `ExtensionContext.Store.CloseableResource` in #1860), so on JUnit < 5.13 the store never closes it.

Combined with this new `owns()` gate, that turns a one-off leak into a per-run leak: on JUnit 5.9–5.12, a Surefire run with `rerunFailingTestsCount` used to reuse the single leaked `Playwright` across all reruns; now every rerun creates a *new* `Playwright` while the previous registry is still never closed, so up to `retries × threads` driver/node processes stay alive for the lifetime of the JVM.

Suggest having `PlaywrightRegistry` implement `ExtensionContext.Store.CloseableResource` as well, so older runtimes actually close the registry and this gate degrades gracefully (5.13+ closes it once via that path and won't double-close).

playwright/src/main/java/com/microsoft/playwright/impl/junit/PlaywrightExtension.java
```diff
@@ -85,13 +88,14 @@ public Object resolveParameter(ParameterContext parameterContext, ExtensionConte
    * @return The Playwright that belongs to the current test.
    */
   public static Playwright getOrCreatePlaywright(ExtensionContext extensionContext) {
+    PlaywrightRegistry registry = PlaywrightRegistry.getOrCreateFor(extensionContext);
     Playwright playwright = threadLocalPlaywright.get();
-    if (playwright != null) {
+    // A previous launcher run on this thread (e.g. a surefire rerun) has already closed its Playwright.
```
yury-s, line 93: nit: this comment describes the fall-through case but sits directly above the *reuse* branch, so it reads as if it justifies the `if` body — easy for a later reader to "fix" by inverting the condition. Moving it below the `if` block, or rewording to something like "reuse only while the current run's registry still owns it", would avoid that.

playwright/src/test/java/com/microsoft/playwright/junit/RerunFixture.java
```diff
@@ -0,0 +1,33 @@
...
+import com.microsoft.playwright.Page;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+// Not picked up by surefire; TestFixturesRerun runs it through the launcher.
+@UsePlaywright
+public class RerunFixture {
+  @Test
+  void usesPage(Page page) {
```
yury-s, line 29: This launches two real browsers serially (once per launcher run) while blocking a JUnit ForkJoinPool worker on `summary.get()`, inside a suite configured with `junit.jupiter.execution.parallel.mode.classes.default = concurrent`. On the 3-vCPU `macos-latest` runner — where the matrix already excludes WebKit for headroom — that's a meaningful addition.

The regression being guarded (`Playwright connection closed`) reproduces without a browser, so injecting `APIRequestContext` here instead of `Page` (or just making a driver round-trip on `Playwright`) would cover the same thing much more cheaply.

yury-s (commented):
The core fix looks right: `PlaywrightRegistry.close()` clears `playwrightList`, so `owns()` correctly returns `false` for a `Playwright` left in the `ThreadLocal` by a previous launcher run, and reuse within a single run is unaffected. I also verified via `help:effective-pom` that importing `junit-bom` first in `<dependencyManagement>` does not override the locally declared entries — `junit-jupiter-api` keeps `provided`, engine/params keep `test`, and `junit-platform-launcher` resolves to 1.14.1 to match Jupiter 5.14.1.

A few comments inline, plus one that I can't anchor because the file isn't in the diff:

**`tools/test-local-installation/pom.xml`** still hand-maintains `junit.platform.version` (1.11.0) alongside `junit.version` (5.11.0). Bumping one without the other yields a mismatched launcher/engine pair that only fails at test time in the docker job. Since this PR already imports `junit-bom` in the root pom, doing the same there would remove the drift.

Develop-KIM:
Thanks for the careful look — all four are in d9a82a6f.

- `PlaywrightRegistry` implements `CloseableResource` again next to `AutoCloseable`, with a comment so it doesn't get cleaned up a second time. I checked both paths: on 5.14.1 the store closes it once through `AutoCloseable` (the `CloseableResource` branch is only reached when AutoCloseable closing is disabled), and on 5.11.0, via the local-installation project, a Playwright captured from the first launcher run now throws "connection closed" afterwards where it previously stayed open.
- Reworded the comment to state the reuse condition and kept it directly on the `if`.
- `RerunFixture` takes an `APIRequestContext` instead of a `Page`. Resolving the parameter already goes through `request().newContext()`, which is enough to trip the bug — without the ownership check the second run still fails with `Playwright connection closed` — and the test now runs in about 1.5s with no browser.
- `tools/test-local-installation/pom.xml` imports `junit-bom`, so `junit.platform.version` is gone; `dependency:tree` there still resolves engine 5.11.0 / launcher 1.11.0. The `-Dgroups=smoke` run in that project is green locally (89 tests).

